PR #18, opened 2019-10-07
fix(pricing): edge cases with division by zero

woyuen:


src/util/pricing.js
```diff
@@ -197,6 +197,7 @@ function getPriceAfterFees (args) {
     takerPrice = _roundPrice(ownerPrice + takerFees)
 
     ownerFeesPercent = getFeesPercent(ownerFees, ownerPrice)
+    takerPrice = _roundPrice(ownerPrice + takerFees)
```
abruere: Please remove the same line `197`.

woyuen: Right, thanks

PR #19, opened 2019-10-07
fix(pricing): takerFeesPercent not applied to base price

### LINKED ISSUE #17: bug(pricing): takerFeesPercent not applied to base price
According to Transaction [API reference](https://docs.api.stelace.com/?version=latest#8e915ba8-0d54-4109-a35d-edbf963c5e2f), `takerFees` should be (transaction) `price * takerFeesPercent` rather than computed inclusively as they currently are, as in `(price + takerFees) * takerFeesPercent`.

woyuen:
According to API reference takerFees should be
(transaction) `price * takerFeesPercent`
rather than computed inclusively
as in `(price + takerFees) * takerFeesPercent`

Closes #17


test/unit/util/pricing.spec.js
```diff
@@ -86,3 +87,28 @@ test('gets the price with custom duration config', (t) => {
     t.deepEqual(getDurationPrice(input), expected)
   })
 })
+
+test('computes price after fees', (t) => {
```
abruere: Maybe `applies takerFeesPercent and ownerFeesPercent to ownerPrice`, more explicit.

woyuen: Thanks

PR #36, opened 2019-10-17
Aggregations

woyuen:
Before only available in Documents and Ratings APIs,
aggregation is now added to Events API .

Same filters are applied to stats and list endpoints.
Moreover `groupBy` and `field` parameters can work with nested properties
such as event `metadata` or `object`,
which is important to aggregate all sort of data (core or custom events).

Improve documents error when `field` is specified while some values are non-number.

Add `object` filter in Event API

### DIFF
--- a/src/util/listQueryBuilder.js
+++ b/src/util/listQueryBuilder.js
@@ -1,4 +1,5 @@
 const _ = require('lodash')
+const createError = require('http-errors')
 
 const { Joi } = require('./validation')
 const { parseArrayValues, getPaginationMeta } = require('./list')
@@ -70,6 +71,14 @@ function getTransformedValues (filters) {
   return values
 }
 
+/**
+ * @param {Object} queryBuilder - Knex.js query builder
+ *   do not use Objection.js functions because `addFiltersToQueryBuilder` is used for list and aggregation.
+ *   However aggregations don't work so great with Objection.js due to `defaultSchema` that is applied to all queries
+ *   even to WITH queries while this is unwanted
+ * @param {Object} filters
+ * @param {Object} transformedValues
+ */
 function addFiltersToQueryBuilder (queryBuilder, filters, transformedValues) {
   Object.keys(filters).forEach(key => {
     const filter = filters[key]
@@ -155,12 +164,13 @@ function checkPaginationConfig (paginationConfig) {
 
 /**
  * Generic query function to address filters, sort and pagination for list endpoints
- * @param {Object}  params.queryBuilder - Objection.js query builder
+ * @param {Object}  params.queryBuilder - Knex.js query builder
  * @param {Object}  [params.filters] - Filters for the query (like select on some IDs)
  * @param {Object}  [params.filters[key]]
- * @param {Object}  [params.filters[key].dbField]
- * @param {Object}  [params.filters[key].type]
- * @param {Object}  [params.filters[key].]
+ * @param {String}  [params.filters[key].dbField]
+ * @param {String}  [params.filters[key].type]
+ * @param {String|Function}  [params.filters[key].transformValue]
+ * @param {String|Function}  [params.filters[key].query]
  *
  * @param {Boolean} [params.paginationActive = true]
  * @param {Object}  [params.paginationConfig]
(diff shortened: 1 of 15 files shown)


abruere:
Thanks! I’ll start reviewing tomorrow.

src/versions/validation/event.js
```diff
@@ -10,10 +10,56 @@ const orderByFields = [
 
 const schemas = {}
 
+const getObjectSchema = (name) => Joi.string().regex(new RegExp(`^${name}\\.\\w+$`))
```
abruere: It is impossible to use nested properties with this regexp. Is it what you expect, and why?

src/versions/validation/event.js
```diff
@@ -10,10 +10,56 @@ const orderByFields = [
...
+    groupBy: groupBySchema.required(),
+    field: fieldSchema,
+    avgPrecision: Joi.number().integer().min(0).default(2),
+
+    // filters
+    id: [Joi.string(), Joi.array().unique().items(Joi.string())],
+    createdDate: getRangeFilter(Joi.string().isoDate()),
+    type: [Joi.string(), Joi.array().unique().items(Joi.string())],
+    objectType: [Joi.string(), Joi.array().unique().items(Joi.string())],
+    objectId: [Joi.string(), Joi.array().unique().items(Joi.string())],
```
abruere: Why not simply coerce using `Joi.array().unique()`?

test/integration/api/event.spec.js
```diff
@@ -14,6 +15,302 @@ test.before(async t => {
...
+ * @param {Object}   params
+ * @param {Object}   params.t - AVA test object
+ * @param {Object}   params.obj - stats object returned by API
+ * @param {String}   params.groupBy
+ * @param {String}   [params.field]
+ * @param {Object[]} params.events
+ * @param {Number}   nbFilteredEvents - check the total number of events
+ * @param {Number}   [avgPrecision = 2] - check the average precision
+ */
+function checkStatsObject ({
```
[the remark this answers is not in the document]
woyuen: That's right.

src/util/listQueryBuilder.js
```diff
@@ -263,6 +273,189 @@ function whereJsonbColumnSupersetOf (queryBuilder, { columnName, data }) {
...
+/**
+ * Transform attribute passed to API to expression that can be used in SQL query
+ * e.g.
+ * `someAttribute` => `"someAttribute"`
+ * `root.property` => `"root"->>'property'`
+ * `root.nested.property` => `("root"->>'nested')::JSON->>'property'`
+ * @param {String} attr
+ * @return {String} expr - SQL expression
+ */
+function getSqlExpression (attr) {
```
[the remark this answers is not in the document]
woyuen: I tested for 3 levels but not with arrays.
I'll add tests for those cases.

src/versions/validation/event.js
woyuen: Good catch. I'll update the regexp so deep nested properties can be used.

src/versions/validation/event.js
woyuen: Right, thanks!

src/util/listQueryBuilder.js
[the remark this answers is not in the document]
abruere: Thanks for switching to `#>>` PostgreSQL jsonb operator.

src/versions/validation/event.js
```diff
@@ -10,7 +10,8 @@ const orderByFields = [
 
 const schemas = {}
 
-const getObjectSchema = (name) => Joi.string().regex(new RegExp(`^${name}\\.\\w+$`))
+// match string value like 'a.b.c' or 'a[0].b.c[2]'
+const getObjectSchema = (name) => Joi.string().regex(new RegExp(`^${name}((\\[(\\d+)\\]|\\.\\w+)*)$`))
```
abruere: Given the complex regex it would be nice to add some pattern [name](https://hapi.dev/family/joi/?v=16.1.7#stringpatternregex-name--options---aliases-regex) to help end user.

```js
Joi.string().pattern(new RegExp(`^${name}((\\[(\\d+)\\]|\\.\\w+)*)$`), {
  name: `accessor string like ${name}.nested.arr[1]`
})
```

woyuen: Thanks!

PR #45, opened 2019-10-21
feat: load an intl polyfill to handle dates

woyuen:
With current version of Node.js (12.7), there is only a partial
support of Intl.
https://nodejs.org/api/intl.html#intl_options_for_building_node_js

For instance, full-text dates are localized only in English.
We need to load a polyfill to handle it,
until Node.js embed `full-icu` by default.

### DIFF
--- a/plugins/email/test/email.spec.js
+++ b/plugins/email/test/email.spec.js
@@ -549,7 +549,7 @@ test('sends an email via a template with ICU content', async (t) => {
     })
     .expect(200)
 
-  const availableDateStr = '2019-03-15T12:45:12.000Z'
+  const availableDateStr = '2019-03-15T18:45:12.000Z'
 
   const payload = {
     name: 'registration',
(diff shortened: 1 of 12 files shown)


plugins/email/test/email.spec.js
```diff
@@ -578,7 +580,12 @@ test('sends an email via a template with ICU content', async (t) => {
   t.true(html.includes('une réduction de 20 %'))
   t.true(html.includes('Votre colis sera livré à votre domicile'))
   t.true(html.includes('à partir du 15/03/2019 (vendredi 15 mars 2019)'))
-  t.true(html.includes('13:45:12'))
+
+  // get hours, minutes and seconds from Date object
+  // because they may differ depending on local time
+  const availableDate = new Date(availableDateStr)
+  const time = `${availableDate.getHours()}:${availableDate.getMinutes()}:${availableDate.getSeconds()}`
```
abruere: According to MDN, [`getX` date methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getHours) depend on local time so how are users supposed to do to not depend on local time of our infrastructure? This is what we should test.

test/integration/api/workflow.spec.js
```diff
@@ -191,6 +191,15 @@ test('creates several single-step Stelace workflows', async (t) => {
...
           endDate: 'new Date(new Date().getTime() + (14 * 24 * 60 * 60 * 1000)).toISOString()',
           isCurrentTest: '_.get(changesRequested, "metadata.singleStepWorkflow")',
+
+          // localized date in Chinese format
+          localizedDate: `
+            new Intl.DateTimeFormat('zh-CH', {
+              day: 'numeric',
+              month : 'long',
+              year : 'numeric'
+            }).format(new Date('2019-02-05'))
```
abruere: Thanks, we’ll need to mention Intl object in Workflow API reference.

plugins/email/test/email.spec.js
woyuen: Thanks, I'll seek methods to allow users to be specify dates independently from our infrastructure.

test/integration/api/workflow.spec.js
abruere: Unless we clearly state in our docs + API reference that our servers use UTC time.

plugins/email/test/email.spec.js
abruere: Unless we clearly state in our docs + API reference that our servers use UTC time.

woyuen: I found a way to specify the timezone in email content and dates will be formatted according to it without any change in ICU messages (no custom format).

I use [`intl-messageformat` custom formatters](https://github.com/formatjs/formatjs/tree/master/packages/intl-messageformat#formatters) and create a custom behaviour if timezone if specified.

https://github.com/stelace/stelace/blob/b40d6f55f4199022469f56ca8f67c2b5fbe617cc/plugins/email/util/i18n.js#L79-L106

If timezone isn't specified, UTC time will be used.

woyuen: Before coming up with this custom formatter solution, I tried the `timeZone` parameter from `Intl.DateTimeFormat`. However our Intl polyfill doesn't support it.
https://github.com/andyearnshaw/Intl.js/blob/4bd382c5df206f328b35ca23579904f603b84f0b/src/12.datetimeformat.js#L198

The date timezone formatter extends the library `intl-messageformat` which is already used by email plugin.

That's right, we should mention in the docs that dates are rendered with UTC time by default.
I think the `timezone` parameter is coherent with what Stelace API offers in Availability and Task API to help users handling timezone issues.

PR #81, opened 2019-11-27
Cron migrate library

woyuen:
Refactor cron logic by using only the library `cron` instead of `cron` and `cron-converter`.


src/util/time.js
```diff
@@ -175,33 +180,32 @@ function isValidCronPattern (pattern) {
  * @param {String} [attrs.timezone='UTC'] - https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
  * @returns {String[]} ISO Dates
  */
-function computeRecurringDates (pattern, { startDate, endDate, timezone = 'UTC' } = {}) {
+function computeRecurringDates (pattern, { startDate, endDate, timezone } = {}) {
```
abruere: Why are you removing this explicit default (and keeping the comment about default above `[attrs.timezone='UTC']`)

src/util/time.js
```diff
@@ -175,33 +180,32 @@ function isValidCronPattern (pattern) {
...
   const dates = []
+  let cronISODate = new Date(new Date(startDate).getTime() - 1) // start from `startDate` minus 1 millisecond
 
   while (continueLoop) {
-    const momentDate = schedule.next()
-
-    const date = momentDate.toISOString()
+    // `_getNextDateFrom` is a private method from `CronTime.prototype`
+    // Please check its availability when upgrading the library `cron`
+    const cronMomentDate = cronTime._getNextDateFrom(cronISODate, timezone)
```
abruere: I don’t see this function in your tests. This would help to detect problems with updates.

test/unit/util/time.spec.js
```diff
@@ -77,24 +118,58 @@ const testOverMonths = (expectedDatesFn, options = {}) => {
...
+
+  testOverMonths(({ m, daysOfMonth, recurringDates, endDate }) => {
+    const expectedDates = getExpectedDates({ m, daysOfMonth, frequency: 'hour', endDate })
+    t.deepEqual(recurringDates, expectedDates)
+  }, 'hour')
+
+  testOverMonths(({ m, daysOfMonth, recurringDates, endDate }) => {
+    const expectedDates = getExpectedDates({ m, daysOfMonth, frequency: 'hour', endDate })
+    t.deepEqual(recurringDates, expectedDates)
+  }, 'hour')
```
abruere: These lines seem duplicated. Maybe you forgot timezone.

src/util/time.js
woyuen: Right. I update the JSDoc.

src/util/time.js
woyuen: Indeed, I'll add a test for it.

test/unit/util/time.spec.js
woyuen: Good catch. The timezone is missing.

src/util/time.js
abruere: You did not tell me why you’re removing the explicit default.

woyuen: Removing the explicit default is a preventive measure against the case where `timezone` is `null`.
This is not to depend on library internal implementation. Currently, passing `null` works because the following condition is falsy:
https://github.com/kelektiv/node-cron/blob/df186541e0ec14a1400c44a053a1c90da98fecbd/lib/cron.js#L27-L33

This measure can probably be removed as `timezone = null` is tested here:
https://github.com/stelace/stelace/blob/2b9ffecc4877e9c73f4fed1148f379af43d01508/test/unit/util/time.spec.js#L165-L168

If so, I'll restore the explicit default value.

abruere: You should think about what the implicit default value is (due to your change) and how it is different from previous explicit default value.

Explicit is better than implicit.

woyuen: My change was to handle the case where `timezone === null` was passed to the function `computeRecurringDates`. So maybe the best way to handle that case is to restore the explicit default value, add the following code into the function.

```js
// @param {String} [attrs.timezone='UTC'] - https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
// if `null`, `timezone` will also be set to the default value ('UTC')
function computeRecurringDates (pattern, { startDate, endDate, timezone = 'UTC' } = {}) {
  if (isNil(timezone)) timezone = 'UTC'
  ...
}
```

abruere: Yes, thanks.

PR #159, opened 2020-02-06
test: create a utility to test webhooks

woyuen:
This utility is exposed to plugins that can use it to perform webhook tests.

### DIFF
--- a/test/webhook-manager.js
+++ b/test/webhook-manager.js
@@ -6,11 +6,13 @@ const { getAccessTokenHeaders } = require('./auth')
 const noop = () => Promise.resolve()
 
 // Utility to test external service webhooks integration into Stelace
-// Two modes of functioning:
-// - test webhooks for real via Ngrok
-// - simulate webhooks by fetching and creating events
-//    (suitable if webhooks cannot be created via API and one cannot have access to Ngrok paid plan
-//     for reserved subdomains)
+// Two operating modes:
+// - Test external webhook integration for real via Ngrok,
+//     as long as the external service lets you create webhooks with some API,
+//     or you have a paid ngrok plan and a stable URL to use for a longlived test webhook
+// - Simulate external webhook by polling external service events
+//    and propagate them to some Stelace endpoint, provided the external service
+//    has some Event API to fetch latest external events.
 class WebhookManager {
   /**
    * @param {Object}   params
@@ -27,11 +29,16 @@ class WebhookManager {
    *   async simulateWebhookSince(lastFetchTimestamp) {
    *     // we recommend to apply deduplication on events
    *     // because events created at the same second can be not available at the same time
-   *     // via Events API
-   *     const events = await exampleSdk.events.list({ gt: lastFetchTimestamp })
+   *
+   *     // timestamps can be expressed in seconds or milliseconds depending on external APIs
+   *     // so use the unit scale that is easier to use in events list filter
+   *     const now = getTimestamp() // implement this function
+   *     const filterParams = lastFetchTimestamp ? { gt: lastFetchTimestamp } : { gt: now }
+   *
+   *     const events = await externalServiceSdk.events.list(filterParams)
    *
    *     for (const event of events) {
-   *       if (processedEventIds[event.id]) continue
+   *       if (processedEventIds[event.id]) continue // via external Events API
    *
    *       await request(t.context.serverUrl)
    *         .post(webhookUrl)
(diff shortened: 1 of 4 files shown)


test/webhook-manager.js
```diff
@@ -0,0 +1,146 @@
...
+const { getAccessTokenHeaders } = require('./auth')
+
+const noop = () => Promise.resolve()
+
+// Utility to test external service webhooks integration into Stelace
+// Two modes of functioning:
+// - test webhooks for real via Ngrok
+// - simulate webhooks by fetching and creating events
+//    (suitable if webhooks cannot be created via API and one cannot have access to Ngrok paid plan
+//     for reserved subdomains)
```
abruere: I would be more precise and make the distinction between external events and Stelace events:

```js
// Two operating modes:
// - Test external webhook integration for real via Ngrok,
//     as long as the external service lets you create webhooks with some API,
//     or you have a paid ngrok plan and a stable URL to use for a longlived test webhook
// - Simulate external webhook by polling external service events
//    and propagate them to some Stelace endpoint, provided the external service
//    has some Event API to fetch latest external events.
```

test/webhook-manager.js
```diff
@@ -0,0 +1,146 @@
...
+   *   poll events from external service and hit Stelace webhook endpoint
+   *   should return the last event timestamp for next fetch
+   *   this timestamp will be injected at the next call of the function `simulateWebhookSince`
+   * @example
+   *   const processedEventIds = {}
+   *
+   *   async simulateWebhookSince(lastFetchTimestamp) {
+   *     // we recommend to apply deduplication on events
+   *     // because events created at the same second can be not available at the same time
+   *     // via Events API
```
abruere: ``` // via external Events API```

And this comment should go with a line having `processedEventIds` below.

test/webhook-manager.js
```diff
@@ -0,0 +1,146 @@
...
+   *     const events = await exampleSdk.events.list({ gt: lastFetchTimestamp })
```
abruere: ```externalServiceSdk.events…```

test/webhook-manager.js
```diff
@@ -0,0 +1,146 @@
...
+   *         .post(webhookUrl)
+   *         .send(event)
+   *         .set({ authorization: `Basic ${basicAuth}` })
+   *         .expect(200)
+   *
+   *       processedEventIds[event.id] = true
+   *     }
+   *
+   *     const timestamps = events.map(e => e.created)
+   *     const maxTimestamp = _.max(timestamps) - 1 // to be sure to fetch processed events
```
abruere: `any unprocessed events`?

And `1` is like a magic number here. Is it supposed to be milliseconds, seconds? Maybe you need some variable name for this safety margin.

I know it’s just an example but this is also a reference for implementation so this has to be easy to grasp.

test/webhook-manager.js
```diff
@@ -0,0 +1,146 @@
...
+   *     return maxTimestamp ? maxTimestamp - 1 : null
+   *   }
+   *
+   * @param {Boolean}  [params.webhookFetchInterval = 2000] - specify the duration in milliseconds for events fetching
```
abruere: `interval in milliseconds between calls to external Events API…`

test/webhook-manager.js
```diff
@@ -0,0 +1,146 @@
...
+    simulateWebhookSince,
+    webhookFetchInterval = 2000,
+    createWebhook,
+    removeWebhook,
+    tunnel
+  }) {
+    this.t = t
+
+    // minus one second to handle cases events are generated during the same second of webhook manager creation
+    this.lastEventTimestamp = getTimestamp() - 1
```
abruere: Same as above: how can you tell it’s seconds and not milliseconds? It depends on the external service.

test/webhook-manager.js
```diff
@@ -0,0 +1,146 @@
...
+      clearInterval(this.simulateWebhookInterval)
+    } else {
+      // https://github.com/bubenshchykov/ngrok#disconnect
+      await ngrok.disconnect(this.tunnelUrl)
+    }
+  }
+
+  // if isWebhookSimulated is false, really wait webhooks events
+  // otherwise, fetch events not retrieved from last time to simulate a real webhook running
+  async waitForEvents (waitDurationMs = 10000) {
```
abruere: It seems it’s just some utility function exposed and comment above is more general advice about using `WehbookManager` than documentation about this function. In which case you should rewrite this comment and move it if needed.

test/webhook-manager.js
```diff
@@ -0,0 +1,146 @@
...
+   *     }
+   *
+   *     const timestamps = events.map(e => e.created)
+   *     const maxTimestamp = _.max(timestamps) - 1 // to be sure to fetch processed events
+   *     return maxTimestamp ? maxTimestamp - 1 : null
+   *   }
+   *
+   * @param {Boolean}  [params.webhookFetchInterval = 2000] - specify the duration in milliseconds for events fetching
+   *   if isWebhookSimulated is true
+   * @param {Function} [params.createWebhook(tunnelUrl)] - can be provided if isWebhookSimulated is false
```
abruere: You could add: `tunnelUrl being the ngrok endpoint to prepend to Stelace API endpoint path.`

I would also provide a full `@example` of `WebhookManager` use with an external webhook (non-simulated).

test/webhook-manager.js
woyuen: Indeed. At initialization, this value is set to `null` now.
The implementation of `simulateWebhookSince` should use the time unit that is easier to use in events list filter.

test/webhook-manager.js
woyuen: Good catch. This comment is obsolete and should be removed.

woyuen:
In fact I realized the security margin was deduced twice (commit before using a variable to name the magic number), so I removed one of subtractions.
https://github.com/stelace/stelace/blob/eb06b4671efe0afac6a1be33df608ae7d6a5a4ff/test/webhook-manager.js#L46-L47

abruere:
Yeah but no reason to remove them both.

LGTM.

PR #295, opened 2020-05-25
Fix: seeding was broken due to update to ava@3

woyuen:


woyuen:
That's a good question.

It makes sense to use directly lifecycle methods instead of relying on AVA.
Later refactor is needed.

PR #297, opened 2020-05-30
refactor: finalize order API

woyuen:
- remove unused code and fix typo
- standardize orderMove and orderLine endpoint URLs
- add a test scenario to simulate a payment

### DIFF
--- a/src/routes/order.js
+++ b/src/routes/order.js
@@ -87,7 +87,7 @@ function init (server, { middlewares, helpers } = {}) {
 
   server.get({
     name: 'orderLine.read',
-    path: '/order_lines/:id'
+    path: '/order-lines/:id'
   }, checkPermissions([
     'orderLine:read',
     'orderLine:read:all'
@@ -105,7 +105,7 @@ function init (server, { middlewares, helpers } = {}) {
 
   server.get({
     name: 'orderMove.read',
-    path: '/order_moves/:id'
+    path: '/order-moves/:id'
   }, checkPermissions([
     'orderMove:read',
     'orderMove:read:all'
@@ -150,7 +150,7 @@ function init (server, { middlewares, helpers } = {}) {
 
   server.post({
     name: 'orderLine.create',
-    path: '/order_lines'
+    path: '/order-lines'
   }, checkPermissions([
     'orderLine:create',
     'orderLine:create:all'
@@ -183,7 +183,7 @@ function init (server, { middlewares, helpers } = {}) {
 
   server.post({
     name: 'orderMove.create',
-    path: '/order_moves'
+    path: '/order-moves'
   }, checkPermissions([
     'orderMove:create',
     'orderMove:create:all'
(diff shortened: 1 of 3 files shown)


src/routes/order.js
```diff
@@ -87,7 +87,7 @@ function init (server, { middlewares, helpers } = {}) {
 
   server.get({
     name: 'orderLine.read',
-    path: '/order_lines/:id'
+    path: '/order-lines/:id'
```
abruere, line 90: Much more consistent with the existing endpoints, thanks.

test/integration/api/order.spec.js
```diff
@@ -656,6 +656,246 @@ test('updates an order move', async (t) => {
...
+  t.is(getCart(user).length, 2)
+
+  // get information before payment
+  const previewedTransactions = await getPreviewedTransactions(user)
+  previewedTransactions.forEach(preview => {
+    t.true(_.isString(preview.assetId))
+    t.true(_.isNumber(preview.quantity))
+    t.true(_.isNumber(preview.unitPrice))
+    t.true(_.isNumber(preview.value))
+  })
```
abruere, line 756: ~Is there some specific reason to test transaction preview objects this here in `order.spec.js`?~

As mentioned in another comment, I suppose it’s because you test the whole shopping cart process, not just the Order API despite the location of this test.

test/integration/api/order.spec.js
```diff
@@ -656,6 +656,246 @@ test('updates an order move', async (t) => {
...
+    await request(t.context.serverUrl)
+      .post('/order-moves')
+      .send({
+        orderId: order.id,
+        payerId: order.payerId,
+        payerAmount: order.amountDue,
+        currency: order.currency,
+      })
+      .set(authorizationHeaders)
+      .expect(200)
```
abruere, line 888: Isn’t this supposed to be transactional in the technical sense?

What if the `/order-moves` request fails after transitions have been applied to transactions and we have an inconsistent state as a result?
I asked about a more integrated sequence on slack but maybe we didn’t spare enough time to think about this.

I don’t promote API complexity, but here we could help Order API user to feel safer and have less work to do to add transitions to every single transaction. Extra parameters might be worth it.

What about some utility parameter (like `transactionTransitionName`, set to a valid transition name) in Order moves API that could automatically add transition to all transactions given transactions `transactionIdsInTransition`, in a **Postgres transaction**.

It’s a feature so it’s not necessary to include such a change in this PR but this could make an issue and a PR on its own.

test/integration/api/order.spec.js
```diff
@@ -656,6 +656,246 @@ test('updates an order move', async (t) => {
   t.is(orderMove.metadata.test, true)
 })
 
+test('simulates a payment process', async (t) => {
```
abruere, line 659: The following test focuses on a precise cart use case (more than the Order API itself) so that you may want to add this keyword to test name, as in `simulates a payment process based on a shopping cart`.

woyuen: Right, I update the test name.

test/integration/api/order.spec.js
woyuen: Yes you're right, it is a transactional operation. Another PR is needed to add this feature.

PR #318, opened 2020-06-15
feat: pagination on workflows and webhooks list

woyuen:
- latest API version serve pagination results for workflows and webhooks list endpionts
- add workflows and webhooks logs endpoints

### DIFF
--- a/src/routes/workflow.js
+++ b/src/routes/workflow.js
@@ -17,9 +17,29 @@ function init (server, { middlewares, helpers } = {}) {
   }, checkPermissions([
     'workflow:list:all'
   ]), wrapAction(async (req, res) => {
-    const params = populateRequesterParams(req)({
+    const fields = [
+      'orderBy',
+      'order',
+      'nbResultsPerPage',
+
+      // cursor pagination
+      'startingAfter',
+      'endingBefore',
+
+      'id',
+      'createdDate',
+      'updatedDate',
+      'event',
+      'active',
+    ]
+
+    const payload = _.pick(req.query, fields)
+
+    let params = populateRequesterParams(req)({
       type: 'list'
     })
+
+    params = Object.assign({}, params, payload)
 
     const result = await requester.send(params)
     return result
--- a/src/routes/workflow.js
+++ b/src/routes/workflow.js
@@ -175,14 +175,13 @@ function init (server, { middlewares, helpers } = {}) {
       'workflowId',
       'eventId',
       'runId',
-      'statusCode',
     ]
 
     const payload = _.pick(req.query, fields)
 
     let params = populateRequesterParams(req)({
       type: 'listLogs',
-      workflowType: req.query.type,
+      logType: req.query.type,
     })
 
     params = Object.assign({}, params, payload)
--- a/src/routes/workflow.js
+++ b/src/routes/workflow.js
@@ -150,6 +150,61 @@ function init (server, { middlewares, helpers } = {}) {
     const result = await requester.send(params)
     return result
   }))
+
+  // //////////// //
+  // WORKFLOW LOG //
+  // //////////// //
+
+  server.get({
+    name: 'workflow.listLogs',
+    path: '/workflow-logs'
+  }, checkPermissions([
+    'workflowLog:list:all'
+  ]), wrapAction(async (req, res) => {
+    const fields = [
+      'orderBy',
+      'order',
+      'nbResultsPerPage',
+
+      // cursor pagination
+      'startingAfter',
+      'endingBefore',
+
+      'id',
+      'createdDate',
+      'workflowId',
+      'eventId',
+      'runId',
+      'statusCode',
+    ]
+
+    const payload = _.pick(req.query, fields)
+
+    let params = populateRequesterParams(req)({
+      type: 'listLogs',
+      workflowType: req.query.type,
+    })
+
+    params = Object.assign({}, params, payload)
+
+    return requester.send(params)
+  }))
+
+  server.get({
+    name: 'workflow.readLog',
+    path: '/workflow-logs/:id'
+  }, checkPermissions([
+    'workflowLog:read:all'
+  ]), wrapAction(async (req, res) => {
+    const { id } = req.params
+
+    const params = populateRequesterParams(req)({
+      type: 'readLog',
+      workflowLogId: id
+    })
+
+    return requester.send(params)
+  }))
 }
 
 function start ({ communication }) {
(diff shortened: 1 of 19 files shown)


src/versions/util.js
```diff
@@ -1,5 +1,6 @@
 // from most recent versions to oldest ones
 const apiVersions = [
+  '2020-06-12', // TODO: adapt the version number
```
abruere: Given the other breaking changes pending, you can probably set this to `2020-07-20` in one month, for now.

src/versions/validation/workflow.js
```diff
@@ -58,6 +89,13 @@ schemas['2019-05-20'].remove = {
 }
 
 const validationVersions = {
+  '2020-06-12': [
+    {
+      target: 'workflow.list',
+      schema: schemas['2020-06-12'].list
```
abruere: Cf. comment about date above

src/routes/workflow.js
```diff
@@ -147,6 +147,58 @@ function init (server, { middlewares, helpers } = {}) {
...
+      'order',
+      'page',
+      'nbResultsPerPage',
+
+      'id',
+      'createdDate',
+      'workflowId',
+      'eventId',
+      'runId',
+      'statusCode',
```
abruere: I’d enable filtering by date rather than `statusCode` that can be missing (e.g. with skipped step).

src/services/workflow.js
```diff
@@ -309,6 +309,109 @@ function start ({ communication, serverPort }) {
...
+      page,
+      nbResultsPerPage,
+
+      id,
+      createdDate,
+      workflowId,
+      eventId,
+      runId,
+      workflowType: type,
+      statusCode,
```
abruere: Please refer to my comments about `type` and `statusCode` above.

src/routes/workflow.js
```diff
@@ -147,6 +147,58 @@ function init (server, { middlewares, helpers } = {}) {
...
+    ]
+
+    const payload = _.pick(req.query, fields)
+
+    let params = populateRequesterParams(req)({
+      type: 'listLogs',
+      workflowType: req.query.type,
```
abruere: `workflowType` is confusing. This is rather `logType`.

src/versions/validation/workflow.js
```diff
@@ -7,13 +7,19 @@ const contextSchema = Joi.array().unique().items(Joi.string()).single()
...
   'name',
   'createdDate',
   'updatedDate',
   'active'
 ]
 
+const workflowLogOrderByFields = [
+  'createdDate',
+  'type',
+  'statusCode',
```
abruere: I would not enable sorting by `statusCode` for reason mentioned above (missing values).

test/integration/api/workflow.spec.js
```diff
@@ -24,6 +24,125 @@ let userWebhookUrl
...
+    permissions: [
+      'workflowLog:list:all',
+      'workflow:create:all',
+      'category:create:all',
+      'entry:create:all',
+      'message:create:all',
+    ]
+  })
+
+  // should create workflows that listen to events that are not triggered by any tests below
```
abruere: Why not integrate assertions related logs to existing workflow tests?

If you can’t and you really need such a full scenario and isolation from other workflow tests you should maybe create another  `workflowLog.spec.js` test file.

test/integration/api/workflow.spec.js
```diff
@@ -1949,6 +2070,93 @@ test('removes a workflow', async (t) => {
...
+
+  t.is(typeof obj, 'object')
+  t.is(obj.nbResults, 1)
+  t.is(obj.nbPages, 1)
+  t.is(obj.page, 1)
+  t.is(typeof obj.nbResultsPerPage, 'number')
+  t.is(obj.results.length, 1)
+})
+
+test('list workflow logs with advanced filters', async (t) => {
```
abruere: `type` filter testing is also needed.

Ideally you could write erroneous code in a workflow to generate a `runError` log.

src/versions/validation/webhook.js
```diff
@@ -34,6 +39,29 @@ schemas['2020-06-12'].list = {
...
+    // pagination
+    page: Joi.number().integer().min(1).default(1),
+    nbResultsPerPage: Joi.number().integer().min(1).max(100).default(DEFAULT_NB_RESULTS_PER_PAGE),
+
+    // filters
+    id: Joi.array().unique().items(Joi.string()).single(),
+    createdDate: getRangeFilter(Joi.string().isoDate()),
+    webhookId: Joi.array().unique().items(Joi.string()).single(),
+    eventId: Joi.array().unique().items(Joi.string()).single(),
+    statusCode: Joi.array().unique().items(Joi.string()).single(),
```
abruere: `statusCode` or `status`?

test/integration/api/webhook.spec.js
```diff
@@ -17,6 +18,86 @@ const { apiVersions } = require('../../../src/versions')
...
+    permissions: [
+      'webhookLog:list:all',
+      'webhook:create:all',
+      'category:create:all',
+      'entry:create:all',
+      'message:create:all',
+    ]
+  })
+
+  // should create webhooks that listen to events that are not triggered by any tests below
```
abruere: Same comment as above with `worflowLog`.

src/routes/workflow.js
woyuen: Ok thanks.

src/routes/workflow.js
woyuen: Indeed it's confusing. I updated the variable name.

test/integration/api/workflow.spec.js
woyuen: I integrated the logs assertion into existing workflows.
I also put workflow/webhook logs into fixtures to be able to test filters in dedicated test.

src/versions/validation/webhook.js
woyuen: Good catch, it's `status`, thanks.

test/integration/api/workflow.spec.js
abruere: Thanks but you are not answering my question, and it’s better to discuss costly changes before making them.

src/services/workflow.js
abruere: This means it doesn’t have an appropriate test.

test/integration/api/workflow.spec.js
woyuen: Indeed, that's not a bad idea to separate workflow/webhook logs into their own test file.
Fixtures are reverted as they were and logs are generated at the start of test suite.

PR #340, opened 2020-07-09
chore(sso): bump built-in OAuth providers’ endpoints

woyuen:
Built-in OAuth connections no longer work after the providers’ endpoints have been updated.
To detect any regression in the future, automated tests are implemented and will run if the environment variable `OAUTH_TEST_CONFIGS` is present.

However only Facebook can be automatically tested in CI:
- Google OAuth process cannot be easily tested with puppeteer because it adapts the authentication process depending if the user is real or a bot.
- Github OAuth process can be only tested in local environment. After submitting credentials at login step, a verification code is sent via email if a new device is detected, even if 2-factor authentication is disabled. And at each test run, a different CircleCI machine is used. Tests for Github OAuth in CircleCI is disabled for now.

Regressions for Github and Google cannot be detected automatically for now.


src/services/authentication.js
```diff
@@ -1683,6 +1689,40 @@ function getSSOLoginUrlClient ({ client, authorizationParams, ssoConnection }) {
   return ssoLoginUrl
 }
 
+function getTokenTypeTransformationsOrder (provider) {
```
abruere, line 1692: This name is not accurate. It can make you think all transformations happen in order. Maybe `getTokenTypeTransformationsToTry`.

src/services/authentication.js
```diff
@@ -1683,6 +1689,40 @@ function getSSOLoginUrlClient ({ client, authorizationParams, ssoConnection }) {
...
+  if (provider === 'facebook') {
+    return [
+      'CAPITALIZE',
```
abruere, line 1695: You could add a comment about why you need to capitalize facebook token type unlike other providers.

test/integration/api/authentication.spec.js
```diff
@@ -15,6 +15,69 @@ const { encodeBase64 } = require('../../../src/util/encoding')
...
+  } catch (err) {
+    return {}
+  }
+})()
+
+const shouldExecuteBuiltInSSOTest = (() => {
+  return typeof !process.env.CIRCLECI ||
+    (
+      process.env.CIRCLECI &&
+        (process.env.CIRCLE_BRANCH !== 'dev' || !process.env.CIRCLE_BRANCH.startsWith('dependabot/'))
```
abruere, line 48: Maybe the body of your commit should also appear here to explain why you exclude these specific branches.

test/integration/api/authentication.spec.js
```diff
@@ -49,8 +112,122 @@ const authorizationServerInfo = {
...
+  })
+
+  const page = await browser.newPage()
+
+  // Chrome on Linux
+  await page.setUserAgent('Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36')
+
+  await executeBrowserScenario({ browser, page })
+
+  await new Promise(resolve => setTimeout(resolve, 1000))
```
abruere, line 151: It sounds like a fragile and magic delay. You need to add some short explanation _at least_ in case test breaks because of this.

test/integration/api/authentication.spec.js
```diff
@@ -49,8 +112,122 @@ const authorizationServerInfo = {
...
+  // with stable URLs
+  // App URL: http://localhost:9461
+  // Authorization callback URL: http://localhost:7645/auth/sso/e8_test/${provider}/callback
+  // Current tested providers: 'github', 'facebook'
+  process.env.SERVER_PORT = 7645
+  const oauthServerPort = 9461
+
+  await before({
+    name: 'authentication',
+    platformId: 8,
```
abruere, line 228: Why platform 8. Can you mention platforms currently used as well (like platform 9?).

test/integration/api/authentication.spec.js
```diff
@@ -49,8 +112,122 @@ const authorizationServerInfo = {
...
+
+  t.deepEqual(user.roles, ['user', 'provider'])
+  t.true(user.platformData.ssoProviders.includes(provider))
+
+  await browser.close()
+}
+
 test.before(async t => {
-  await before({ name: 'authentication' })(t)
+  // set fixed values to be able to test OAuth authentication with external providers
```
abruere, line 218: I believe you specifically mean “fixed **port numbers**”?

test/integration/api/authentication.spec.js
```diff
@@ -899,6 +1076,140 @@ test.serial('performs an OpenID authentication', async (t) => {
...
+        await page.type(usernameSelector, username)
+        await page.type('input[type="password"]', password)
+        await page.click('input[type="submit"]')
+
+        // authorize form is displayed the first time
+        const authorizeButtonSelector = 'button[name="authorize"]'
+        const authorizeButtonExists = await checkElementVisible(page, authorizeButtonSelector)
+        if (authorizeButtonExists) {
+          // wait a little bit before clicking the button as it can be disabled at first
+          await new Promise(resolve => setTimeout(resolve, 1000))
```
abruere, line 1118: I suppose there is no mean to wait for a given state of the button using puppeteer? (like what you can do to wait for an element to show up)

test/integration/api/authentication.spec.js
```diff
@@ -1077,6 +1077,14 @@ test.serial('performs an OpenID authentication', async (t) => {
 })
 
 if (shouldExecuteBuiltInSSOTest) {
+  // Please note Github may apply a verification code at login when a new detected device
+  // even if 2-factor authentication is disabled
+  // https://github.blog/changelog/2019-07-01-verified-devices
+  // https://github.community/t/disable-remove-email-device-verification-prompt-on-login-not-the-2fa/2333
+
+  // You should login to Github with your test account credentials (username, password) before running
+  // this test in local environment
```
abruere, line 1086: Do you mean in a specific browser (given that tests OAuth2 tests run in Puppeteer)?

test/integration/api/authentication.spec.js
```diff
@@ -1077,6 +1077,14 @@ test.serial('performs an OpenID authentication', async (t) => {
...
+  // This security measure makes CI testing really hard
```
abruere, line 1087: // This security measure makes CI testing really hard
// and `github` provider should generally be omitted in `OAUTH_TEST_CONFIGS`

(same in your commit message body).

test/integration/api/authentication.spec.js
woyuen: Yes thanks, I'll add an explanation.

test/integration/api/authentication.spec.js
woyuen: Currently besides platform 9, the platform 1 is reserved for instant seed and 2 for api key tests.

https://github.com/stelace/stelace/blob/654ea5afc38bc4f99595edce01bde6d4e433d51c/scripts/instantData.js#L15
https://github.com/stelace/stelace/blob/654ea5afc38bc4f99595edce01bde6d4e433d51c/test/integration/api/apiKey.spec.js#L15

Is it better if I set it as platform 3 ?

test/integration/api/authentication.spec.js
woyuen: I'll try to make it more robust with `.waitForFunction()` which can evaluate an expression (element visibility and not disabled if it's a button).
https://github.com/puppeteer/puppeteer/blob/v5.0.0/docs/api.md#framewaitforfunctionpagefunction-options-args

test/integration/api/authentication.spec.js
woyuen: I didn't encounter issues after sign in with Chromium. I suppose the new device verification detects a mix of user agent and ip address.
For local environment, signing in with any browser should pass the verification step, so testing puppeteer will be permitted.

test/integration/api/authentication.spec.js
abruere: Yes please, it’s easier to keep track of used test platforms this way.

PR #349, opened 2020-07-17
perf: add TimescaleDB extension to PostgreSQL

woyuen:
Time-series tables are intended to be very large due to their nature (e.g. events, webhook/workflow logs).

Even with indices optimization and partitioning, PostgreSQL can't beat TimescaleDB in terms of performance (the latter is optimized for time-series):
https://blog.timescale.com/blog/timescaledb-vs-6a696248104e

The good point is that TimescaleDB is an extension of PostgreSQL so it will benefit from existing tools and the codebase migration won't be too massive.
Time-series tables will be transformed into Timescale hypertables.

Knex migration file is reset to take into account new tables structure.
For performance purposes, missing indexes are added for most of list filters to speed queries.
This is also the opportunity to create multi-columns indexes (such as `(createdDate, id)`) to prepare upcoming cursor-based pagination, which is more scalable than the current offset-limit pagination.

To drastically decrease storage usage, hypertable compression (https://docs.timescale.com/latest/using-timescaledb/compression) will be activated for data older than 30 days.
During our tests with 22M rows table, usage size was divided by 3 (64GB to 22GB).
This is a breaking change: compressed individual rows retrieval won't be allowed anymore because filtering on compressed data is slow. Indexes won't be used, so a full scan is needed.

Event stats endpoint will be deprecated in favor of history endpoint because it's really hard to optimize, due to the flexibility offered by the endpoint, when the table becomes large.
To gain performance on aggregation, we'll use continuous aggregates (https://docs.timescale.com/latest/using-timescaledb/continuous-aggregates) to pre-aggregate data.
Continuous aggregates are an improved version of PostgreSQL materialized views, including features such as automated data synchronization.


src/services/event.js
```diff
@@ -215,7 +224,10 @@ function start ({ communication }) {
 
     const eventId = req.eventId
 
-    const event = await Event.query().findById(eventId)
+    const minCreatedDate = getRetentionLimitDate()
+
+    // without filter, compressed chunk will be queried so the response will be long
```
abruere: Do you mean:

```js
// without this filter, compress chunks would be queried…
```

src/util/listQueryBuilder.js
```diff
@@ -50,22 +52,24 @@ function getTransformedValues (filters) {
...
-    if (typeof transformValue === 'undefined') {
+    if (_.isUndefined(transformValue)) {
       newValue = value
-    } else if (typeof transformValue === 'string') {
+    } else if (_.isString(transformValue)) {
       if (transformValue === 'array') {
         newValue = parseArrayValues(value)
       }
-    } else if (typeof transformValue === 'function') {
+    } else if (_.isFunction(transformValue)) {
```
abruere, line 65: Any reason or specific bug to start using lodash functions now or just refactoring along the way?

src/util/timeSeries.js
```diff
@@ -0,0 +1,18 @@
+const { computeDate, getPureDate } = require('./time')
+const ms = require('ms')
+
+const retentionLogDuration = '31 days'
+
+function getRetentionLimitDate () {
+  // round to pure date (0h0m0s) to have a relative stable value to display if there is any error value
```
abruere: `relatively stable`?

src/util/timeSeries.js
```diff
@@ -0,0 +1,18 @@
...
+  // example: from 2020-01-31 (included) to 2020-02-01 (excluded) will have 2020-01-01
+  // as retention limit date if the retention duration is '31 days'
+  const today = getPureDate(new Date().toISOString())
```
abruere: Maybe it’s time to give this function a more explicit name like `getRoundedDate`.

test/integration/api/event.spec.js
```diff
@@ -323,6 +323,76 @@ test('aggregation works even if the specified nested property does not exist', a
...
+
+  await request(t.context.serverUrl)
+    .get(`/events/stats?groupBy=${groupBy}&createdDate=${invalidMinCreatedDate}`)
+    .set(authorizationHeaders)
+    .expect(400)
+
+  const { body: obj } = await request(t.context.serverUrl)
+    .get(`/events/stats?groupBy=${groupBy}&createdDate[gte]=${validMinCreatedDate}`)
+    .set(authorizationHeaders)
+    .expect(200)
```
abruere, line 581: Maybe you could check the content of error message (some keword using regular expression as in other tests and test files)?

test/integration/api/event.spec.js
```diff
@@ -464,6 +534,52 @@ test('list events with filters', async (t) => {
...
+
+  t.is(obj.results.length, obj.nbResults)
+  obj.results.forEach(event => {
+    t.true(event.createdDate >= validMinCreatedDate)
+  })
+})
+
+test('list events only within retention log period', async (t) => {
```
abruere: Maybe you can add `by default`.

test/integration/api/event.spec.js
```diff
@@ -464,6 +534,52 @@ test('list events with filters', async (t) => {
...
+
+  await request(t.context.serverUrl)
+    .get(`/events?createdDate=${invalidMinCreatedDate}`)
+    .set(authorizationHeaders)
+    .expect(400)
+
+  const { body: obj } = await request(t.context.serverUrl)
+    .get(`/events?createdDate[gte]=${validMinCreatedDate}`)
+    .set(authorizationHeaders)
+    .expect(200)
```
abruere, line 783: Same about checking error message.

src/util/listQueryBuilder.js
```diff
@@ -83,25 +87,41 @@ function getTransformedValues (filters) {
...
     const filter = filters[key]
-    const { dbField, value, query } = filter
+    const { dbField, value, query, minValue } = filter
     const transformedValue = transformedValues[key]
 
-    // skip the filter if the value is not defined
-    if (typeof value === 'undefined') {
-      return
-    }
+    // skip the filter if the value is not defined,
```
abruere: typo: trailing comma

src/util/listQueryBuilder.js
```diff
@@ -83,25 +87,41 @@ function getTransformedValues (filters) {
...
+    if (_.isUndefined(query)) {
       queryBuilder.where(dbField, transformedValue)
-    } else if (typeof query === 'string') {
+    } else if (_.isString(query)) {
       if (query === 'range') {
         checkRangeFilter(transformedValue, key)
 
+        const isSingleValue = !_.isPlainObject(transformedValue)
+
+        if (!_.isUndefined(minValue)) {
```
abruere: Can you explain why you specifically test `undefined` and not just truthiness `if (minValue)`?

migrations/knex/20200713111032_init.js
```diff
@@ -43,11 +45,11 @@ exports.up = async (knex) => {
...
-    table.index('assetId')
-    table.index('ownerId')
-    table.index('takerId')
-    table.index('emitterId')
-    table.index('receiverId')
+    table.index('assetId', 'assessment_assetId_index')
+    table.index('ownerId', 'assessment_ownerId_index')
+    table.index('takerId', 'assessment_takerId_index')
+    table.index('emitterId', 'assessment_emitterId_index')
+    table.index('receiverId', 'assessment_receiverId_index')
```
abruere, line 60: It would make review much safer if you split such index renaming in its own commit.

migrations/knex/20200713111032_init.js
```diff
@@ -225,8 +226,9 @@ exports.up = async (knex) => {
   })
 
   await knex.schema.createTable('event', table => {
-    table.string('id').primary()
```
abruere, line 205: It would be helpful for future reader to specify why you remove this in commit message.

migrations/knex/20200713111032_init.js
```diff
@@ -225,8 +226,9 @@ exports.up = async (knex) => {
...
+    table.string('id')
     table.string('createdDate', 24)
+    table.timestamp('createdTimestamp', { precision: 3 })
```
abruere: You may want to explain why you pick `3`.

migrations/util/timescaleDB.js
```diff
@@ -0,0 +1,13 @@
+// https://docs.timescale.com/latest/using-timescaledb/hypertables
+function createHypertable (schema, table, { column = 'createdTimestamp', migrateData = false } = {}) {
+  return `
+    SELECT public.create_hypertable(
+      '${schema}."${table}"', '${column}'
+      ${migrateData ? ', migrate_data => true' : ''}
```
abruere, line 14: Maybe you want to comment on `migrate_data` and performance?

src/models/Event.js
```diff
@@ -122,8 +125,13 @@ class Event extends Base {
...
     const now = new Date().toISOString()
 
-    if (!this.createdDate) {
+    if (!this.createdDate && !this.createdTimestamp) {
       this.createdDate = now
+      this.createdTimestamp = now
+    } else if (!this.createdDate) {
+      this.createdDate = this.createdTimestamp
+    } else if (!this.createdTimestamp) {
+      this.createdTimestamp = this.createdDate
```
abruere, line 134: Conversion is automatic, right?

About createdDate, I wondered if we shouldn’t set this in database (using some function for instance) rather than in server to ensure insertion order is the same as ordering by date (could not be the case with concurrent insertions by several server instances).

It would be safer in case we want to use PostgreSQL `BRIN` indexes in the future, potentially with non hypertables if it conflicts with TimescaleDB in some way (data ordering).

abruere (commented):
Thanks! This is only the first pass (first commits only, before `getHistory` endpoint).

About compression, could you give an order of magnitude of compression and specify this in PR description and in commit `perf!: add compression policy for hypertables` as well?
In this very commit, maybe you want to explain that you restrict query date range for performance reasons (commit message lets you think it’s not feasible but doesn’t tell why).
Plus other changes discussed on slack.

src/services/event.js
woyuen: Yes, thanks.

src/util/listQueryBuilder.js
woyuen: That's just refactoring.

src/util/timeSeries.js
woyuen: The idea was to have a retention limit date that remains the same for one-day duration, instead of subtracting 30 days duration from the current time (which would return a different value each time).
I'll update the comment as:
`round to pure date (0h0m0s) to return the same value for one-day duration...`

src/util/listQueryBuilder.js
woyuen: This range filter can be used to filter any type.
https://github.com/stelace/stelace/blob/0aa12cc5066778bbd6cc893d9b8da497ce134451/src/util/validation.js#L101-L109

For the specific case of number, an allowed value is `0`, therefore I prefer to test if the value isn't `undefined`.

migrations/knex/20200713111032_init.js
woyuen: Indeed, I'll split into its own commit.

migrations/knex/20200713111032_init.js
woyuen: Thanks, I'll add an explanation to the commit message.

migrations/knex/20200713111032_init.js
woyuen: Indeed, I'll create a variable at the top of the file for `const millisecondPrecision = 3`.

migrations/util/timescaleDB.js
woyuen: Yes, thanks. I'll add a comment about it.

src/models/Event.js
woyuen: Yes the conversation is automatic.

Good idea. Maybe not for the model event because we set the `createdDate` manually to match the targeted resource date.

Example for user creation event : https://github.com/stelace/stelace/blob/0aa12cc5066778bbd6cc893d9b8da497ce134451/src/services/user.js#L874-L879

 Any other model can benefit from the PostgreSQL function. I'll add this function.

woyuen:
Thanks for this first review.

Yes I'll add a description on compression ratio and the performance reason behind the date filter restriction.

src/util/timeSeries.js
abruere: I was just fixing your typo `relative stable` -> `relatively stable`.

migrations/knex/20200713111032_init.js
abruere: It could make you think it’s 3 milliseconds, not 10^(-3) second as it seems to be.
So maybe `const timestampPrecision = 3 // 10E-3 = 1 millisecond`

src/models/Event.js
abruere: It’s not so useful if we don’t use it for Event that is the most important table for timescaleDB.
Moreover, lately you said on slack you probably wouldn’t use BRIN indexes, so why do you still want to create the function?

woyuen: That's right. Then there's no need to create this function.

src/util/time.js
```diff
@@ -140,16 +117,23 @@ function computeDate (isoDate, duration) {
 }
 
 /**
- * Round the date, by default to the nearest minute
+ * Round the date, by default to the inferior date UTC with 0h0m0s
```
abruere: This behavior is inconsistent and doesn’t match the name of the function. Why not round in all cases?

src/util/time.js
```diff
@@ -140,16 +117,23 @@ function computeDate (isoDate, duration) {
...
-function getRoundedDate (date, nbMinutes = 1) {
+function getRoundedDate (date, { nbMinutes } = {}) {
   const d = isDate(date) ? date : new Date(date)
 
-  const ms = nbMinutes * 60 * 1000
-  const roundedDate = new Date(Math.round(d.getTime() / ms) * ms)
-  return roundedDate.toISOString()
+  const roundToDate = _.isUndefined(nbMinutes)
+
+  if (roundToDate) {
```
abruere, line 141: Even this boolean says that you’re expected to round, not take floor.

src/util/timeSeries.js
```diff
@@ -1,13 +1,13 @@
-const { computeDate, getPureDate } = require('./time')
+const { computeDate, getRoundedDate } = require('./time')
 const ms = require('ms')
 
 const retentionLogDuration = '31 days'
 
 function getRetentionLimitDate () {
-  // round to pure date (0h0m0s) to have a relative stable value to display if there is any error value
+  // round to inferior UTC date with 0h0m0s to have a relative stable value to display if there is any error value
```
abruere: Here indeed you want to take floor rather than rounded value. Maybe you can just subtract 12 hours or probably better, create a new function to truncate dates.
Please test appropriately anyway.

test/unit/util/time.spec.js
```diff
@@ -43,14 +43,19 @@ test('computes a date based on an object duration', (t) => {
 })
 
 test('get rounded date', (t) => {
-  // round to nearest minute
+  // round to inferior date UTC
   t.is(getRoundedDate('2018-01-01T00:00:00.000Z'), '2018-01-01T00:00:00.000Z')
-  t.is(getRoundedDate('2018-01-01T00:00:00.001Z'), '2018-01-01T00:00:00.000Z')
-  t.is(getRoundedDate('2018-01-01T00:00:50.000Z'), '2018-01-01T00:01:00.000Z')
+  t.is(getRoundedDate('2018-01-01T12:45:50.000Z'), '2018-01-01T00:00:00.000Z')
+  t.is(getRoundedDate('2018-01-01T19:00:00.000Z'), '2018-01-01T00:00:00.000Z')
```
abruere: This is the confusing use of the function supposed to round dates, not to truncate them.

abruere (changes requested):
Specific review of `getRoundedDate` refactoring.

src/services/event.js
```diff
@@ -321,7 +321,7 @@ function start ({ communication }) {
 
     const minCreatedDate = getRetentionLimitDate()
 
-    // without filter, compressed chunk will be queried so the response will be long
+    // without this filter, compressed chunk will be queried so the response will be long
```
abruere: You forgot `will` -> `would` (please replace both occurrences by the way).

.circleci/config.yml
```diff
@@ -200,7 +200,7 @@ jobs:
               /tmp/test-results/unit-test-results.xml
       - run:
           name: Run integration tests with API endpoints
-          no_output_timeout: 12m
+          no_output_timeout: 20m
```
abruere: We should have to make such tweaks on a regular basis, and 20m is too long since our CI tests are split across 4 containers. If a test does not end in say 8 minutes (4 minutes for all other tests) it probably won’t in 16 minutes either.

Do you have contradictory evidence? Otherwise you probably should proceed by elimination and spot which test cannot end and reaches timeout.
Some help: in this unrelated PR (dependabot), tests ran out in container 4: https://circleci.com/gh/stelace/stelace/1968. You can have a look at artifacts. The culprit is _not_ in artifacts of containers 1 to 3.
It _could_ be in artifacts of container but could be missing as well due to timeout.

abruere (commented):
About test timeout change.

src/util/time.js
woyuen: Yes right, I'll change that.

src/util/time.js
woyuen: Yes, that's right.

src/util/timeSeries.js
woyuen: Yes 'truncateDate' is a more appropriate name for this function.
The objective to truncating instead of rounding is to have at least 30 complete days of retention log.

test/unit/util/time.spec.js
woyuen: Indeed, I'll update the behavior.

.circleci/config.yml
woyuen: Thanks, that's true that tests are run in parallel and each container won't run the total number of tests, like in local environment.
I'll find the culprit test that throws this timeout error.

test/unit/util/time.spec.js
```diff
@@ -42,11 +43,19 @@ test('computes a date based on an object duration', (t) => {
...
+
 test('get rounded date', (t) => {
-  // round to inferior date UTC
+  // round to nearest date UTC
   t.is(getRoundedDate('2018-01-01T00:00:00.000Z'), '2018-01-01T00:00:00.000Z')
-  t.is(getRoundedDate('2018-01-01T12:45:50.000Z'), '2018-01-01T00:00:00.000Z')
-  t.is(getRoundedDate('2018-01-01T19:00:00.000Z'), '2018-01-01T00:00:00.000Z')
+  t.is(getRoundedDate('2018-01-01T05:00:00.000Z'), '2018-01-01T00:00:00.000Z')
+  t.is(getRoundedDate('2018-01-01T12:45:50.000Z'), '2018-01-02T00:00:00.000Z')
+  t.is(getRoundedDate('2018-01-01T19:00:00.000Z'), '2018-01-02T00:00:00.000Z')
```
abruere, line 59: Thanks, you should probably test `12:00:00` rounding too.

.circleci/config.yml
woyuen: It seems the timeout error is related to authentication tests with `puppeteer` browsing.
In rare occasions, the error `Protocol error (Runtime.callFunctionOn): Target closed.` will occur when waiting for a selector, like here:
https://github.com/stelace/stelace/blob/17fca25cba500f12470789cf862eb3f8e80067cb/test/integration/api/authentication.spec.js#L624

The problem is that error isn't thrown and the test suite never finishes, which explains the CircleCI exceeded no output timeout.

I tried to add the [recommended flag](https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#setting-up-chrome-linux-sandbox) `--disable-setuid-sandbox`, in addition of already provided flags `--disable-dev-shm-usage`, `--no-sandbox`, but that didn't help.
https://github.com/stelace/stelace/blob/17fca25cba500f12470789cf862eb3f8e80067cb/test/integration/api/authentication.spec.js#L614-L617

As explained in this [Stack Overflow post](https://stackoverflow.com/questions/51629151/puppeteer-protocol-error-page-navigate-target-closed/51989560), it seems the browser has an initialization issue that can happen from time to time.

For now, I have no better solution that re-triggering CircleCI workflows after such failing no output timeout errors.

abruere: Thank you for investigating. You can probably turn this into a standalone issue (github button `...`, top right of your message).

src/services/event.js
```diff
@@ -16,6 +16,93 @@ function start ({ communication }) {
...
+        emitter: {
+          dbField: 'emitter',
+          value: emitter
+        },
+        emitterIds: {
+          dbField: 'emitterId',
+          value: emitterId,
+          transformValue: 'array',
+          query: 'inList'
+        }
```
abruere, line 97: How’s performance with fields not indexed (such as emitter[Id])?

src/util/listQueryBuilder.js
```diff
@@ -451,6 +447,119 @@ async function performAggregationQuery ({
...
+
+  const { orderBy, order } = orderConfig
+  queryBuilder.orderBy(orderBy, order)
+
+  const { page, nbResultsPerPage } = paginationConfig
+
+  const allResults = await queryBuilder
+
+  // To avoid query complexity, data retrieval for pagination aren't split into results query and count query.
+  // TODO: improve it with the incoming cursor-based pagination
```
abruere, line 629: I see you’re using `count()` above, which would mean that you get both results and count according to your comment?
Why not do the same for other models?

src/versions/validation/event.js
```diff
@@ -101,6 +123,10 @@ schemas['2019-05-20'].create = {
 
 const validationVersions = {
   '2019-05-20': [
+    {
+      target: 'event.getHistory',
+      schema: schemas['2019-05-20'].getHistory
```
abruere: Shouldn’t you add this endpoint to future version?

test/util.js
```diff
@@ -209,3 +210,75 @@ function checkStatsObject ({
...
 }
+
+/**
+ * @param {Object}   params
+ * @param {Object}   params.t - AVA test object
+ * @param {Object}   params.obj - stats object returned by API
+ * @param {String}   params.groupBy
+ * @param {Object[]} params.results
+ * @param {String}   [order = 'desc']
+ * @param {Function} [additionalResultCheckFn] - if defined, will perform additional check on `result`
```
abruere, line 222: Imprecise.
It seems it is `on each item of results array`.

src/util/listQueryBuilder.js
```diff
@@ -451,6 +447,119 @@ async function performAggregationQuery ({
...
+  checkPaginationConfig(paginationConfig)
+
+  if (!['hour', 'day', 'month'].includes(groupBy)) throw new Error(`Invalid groupBy value: ${groupBy}`)
+
+  const intervals = {
+    hour: '1 hour',
+    day: '1 day',
+
+    // TimescaleDB doesn't support month interval (https://github.com/timescale/timescaledb/issues/414)
+    month: '30 days'
```
abruere, line 528: This is embarassing as it was on of the requirements for the migration.
Please ensure it’s specified in the updated API reference (have the draft ready before merging).

src/util/listQueryBuilder.js
```diff
@@ -471,7 +580,24 @@ function getSqlExpression (attr) {
...
+
+  const isSingleValue = !_.isPlainObject(rangeOrValue)
+  if (isSingleValue) return rangeOrValue
+
+  const rangeValues = [rangeOrValue.gt, rangeOrValue.gte]
+  const [minValue] = rangeValues.sort((a, b) => {
+    if (_.isUndefined(a)) return 1
+    else if (_.isUndefined(b)) return -1
+    else return a < b ? -1 : 1
+  })
```
abruere: Very contrived, you can make this much simpler to read (e.g without sort function).

src/util/listQueryBuilder.js
```diff
@@ -451,6 +447,119 @@ async function performAggregationQuery ({
   return paginationMeta
 }
 
+/**
+ * Generic query function to address filters, sort and pagination for stats endpoints
```
abruere: Do you write `stats` on purpose to include deprecated stats endpoint, or should you replace with `history endpoint`?

src/database.js
```diff
@@ -40,16 +40,15 @@ async function dropColumnsIfExist (knex, tableName, columns) {
...
-  const knex = Knex(params)
+/**
+ * @param {Object}  connection - provide `connection` or `knex`
+ * @param {Object}  knex - Knex.js query builder
+ * @param {String}  schema
+ * @param {String}  [destroyKnex = true] - if false, knex isn't destroyed so it can be reused
+ * @return {Object|Undefined} returns knex if `destroyKnex` is false
+ */
+async function createSchema ({ connection, schema, knex, destroyKnex = true }) {
+  if (!knex) knex = getKnex({ connection, schema })
```
abruere, line 51: Why destroy by default if the function is able to create its own short-lived knex instance?

src/database.js
```diff
@@ -97,6 +89,49 @@ async function dropSchema ({ connection, schema, cascade = false, destroyKnex =
...
 
+/**
+ * @param {Object}  connection - provide `connection` or `knex`
+ * @param {Object}  knex - Knex.js query builder
+ * @param {String}  schema
+ * @param {String}  [destroyKnex = true] - if false, knex isn't destroyed so it can be reused
+ * @return {Object|Undefined} returns knex if `destroyKnex` is false
+ */
+async function dropSchemaViews ({ connection, schema, knex, destroyKnex = true }) {
+  if (!knex) knex = getKnex({ connection, schema })
```
abruere, line 100: Same.

src/util/listQueryBuilder.js
```diff
@@ -520,30 +524,53 @@ async function performHistoryQuery ({
...
-  queryBuilder.orderBy(orderBy, order)
+    const { orderBy, order } = orderConfig
+    queryBuilder.orderBy(orderBy, order)
 
-  const { page, nbResultsPerPage } = paginationConfig
+    allResults = await queryBuilder
+  } else {
+    // `AT TIME ZONE 'UTC'` statement is important to add to have the type 'timestamptz'
+    // otherwise the type 'timestamp' is returned and the date won't be UTC
+    const select = knex.raw(`public.time_bucket(INTERVAL '${interval}', ??) AT TIME ZONE 'UTC' as ??`, [timeColumn, groupBy])
```
abruere: Not very DRY.
And can you explain why time_bucket is not used when `onlyTimeFilter && view` above? Already embedded in continuous aggregate? Not obvious when you read the code.

migrations/util/timescaleDB.js
```diff
@@ -30,13 +30,15 @@ function createContinuousAggregate ({
   interval,
   timeBucketLabel,
   column = 'createdTimestamp',
+  secondaryColumn
```
abruere: Don’t you think we may add other columns in the future?
You’re probably right if you don’t, as having `secondaryColumns` would probably break the YAGNI rule. But it never hurts to plan a bit and explain your choices.

We need to ensure compressed rows have enough data in TimescaleDB and indeed there shouldn’t be too many dimensions to keep effective compression. You should probably specify this in your commit message, or comment TimescaleDB code.

migrations/util/timescaleDB.js
```diff
@@ -29,16 +29,42 @@ function createContinuousAggregate ({
...
+  // won't be materialized (`refresh_lag` + `bucket_width`)
+
+  // We recommend to set:
+  // - '1 day' for `refresh_lag` and `refresh_interval` if `bucket_width` superior to '1 day'
+  // - the same value of `bucket_width` if it's inferior to '1 day'
+  refreshLag = '1 day',
+  refreshInterval = '1 day',
+
+  // Any modifications on rows whose date is beyond this period won't be propagated to continuous aggregate
+  // This will enable rows removal (for storage purpose) without impacting continuous aggregates
```
abruere: `(for storage purpose)` -> `(to save storage space)`

abruere (commented):
And now, `getHistory`.

src/services/event.js
woyuen: For event, only `emitter` isn't indexed because there are only 3 currently allowed values (core, task, custom).
The query planner won't choose to use the index on that column even if it exists.

Concerning the response time, the test platform with 3M on non-compressed rows (filter on columns are only allowed within the retention log period), it's pretty good (<100ms). That can be explained by the fact the different emitter values are evenly distributed.

src/util/listQueryBuilder.js
woyuen: Actually this `count()` is the data we want to retrieve in rows, not the pagination count.
To have the pagination count, we need to do a `count` of `count`, which I chose not to do because of incoming cursor pagination which doesn't perform this count request.

src/versions/validation/event.js
woyuen: Yes right, I'll add this endpoint to a new version.

test/util.js
woyuen: Thanks, I'll update the comment.

src/util/listQueryBuilder.js
woyuen: Good catch, there's a typo. It should be `history endpoint`.

src/database.js
woyuen: Right, I'll change that.

src/util/listQueryBuilder.js
woyuen: I'll add a comment on that point.
The `time_bucket` is already embedded in continuous aggregate definition indeed.
https://github.com/stelace/stelace/blob/1306c1b13e162567e60f7f04668507061d4f1612/migrations/util/timescaleDB.js#L62-L67

migrations/util/timescaleDB.js
woyuen: I considered passing an array of columns but kept only one secondary column because the number of dimensions will quickly increase the number of rows if the different dimensions have many possible values.

I'll add an explanation for this choice, thanks.

src/util/listQueryBuilder.js
```diff
@@ -623,14 +623,13 @@ function getMinRangeValue (rangeOrValue) {
...
-  const rangeValues = [rangeOrValue.gt, rangeOrValue.gte]
-  const [minValue] = rangeValues.sort((a, b) => {
-    if (_.isUndefined(a)) return 1
-    else if (_.isUndefined(b)) return -1
-    else return a < b ? -1 : 1
-  })
-
-  return minValue
+  if (!_.isUndefined(rangeOrValue.gt) && !_.isUndefined(rangeOrValue.gte)) {
+    return rangeOrValue.gt < rangeOrValue.gte ? rangeOrValue.gt : rangeOrValue.gte
```
abruere: You’re not throwing but **implicitly** (not so great) merging the two parameters like an `OR` (lowest of two min values).

Does the same strategy applies to all current list endpoints? It does not seem to be properly documented.
When you have both `gt|gte` and `lt|lte` filters in list endpoints, those filters are joined with `AND` to make sense (filter values between `gt` and `lt`).
Then such implicit `OR` merging can be confusing.

src/database.js
abruere: Thanks.
By the way it’s better for dev experience to name boolean parameters after their optional behavior, so you know what you’re doing when using or omitting them.

`destroyKnex` made you think it’s an option (`false` by default), whereas knex was destroyed (`true`) by default.

abruere: The problem is you probably omitted to mention you were destroying knex to avoid memory leaks by use of `createSchema` and `dropSchema` without destroying knex when not using. Am I right?

In such case, maybe the fix was just to rename `destroyKnex` (that was true by default) into `returnKnex` (false by default).

Anyway, you can reuse this commit instead of creating a new one, just explain your plan here.

src/util/listQueryBuilder.js
```diff
@@ -529,57 +529,64 @@ async function performHistoryQuery ({
...
+  const useContinuousAggregate = useOnlyNonTimeRestrictedFilters && view
+
+  const select = (() => {
+    const selectParams = []
+    let selectQuery = ''
+
+    if (useContinuousAggregate) {
+      // Please consult continuous aggregate definition in `migrations/util/timescaleDB.js`
+      // it already embeds `time_bucket` function, thus no need to specify it via request
+      // only reference it via `groupBy` value (hour, day or month)
```
abruere: Explicit variable `useContinuousAggregate` and this comment are welcome, but the string concatenation makes it hard to see the final result.

src/util/listQueryBuilder.js
```diff
@@ -529,57 +529,64 @@ async function performHistoryQuery ({
...
-      const transformedValues = getTransformedValues(newFilters)
-      addFiltersToQueryBuilder(queryBuilder, newFilters, transformedValues)
+    selectQuery += ', '
 
-      if (secondaryFilter) queryBuilder.groupBy(groupBy)
+    if (useContinuousAggregate) {
+      if (secondaryFilter) selectQuery += 'sum(count)::REAL as count'
+      else selectQuery += 'count'
+    } else {
+      selectQuery += 'count(*)'
```
abruere, line 596: This is the most confusing part in your string concatenation.

It’s unclear why use `count` in one case and `count(*)` in the other. This relates to use of `knex.count()` you removed below.

abruere (commented):
Thanks, here is my review of last changes.

src/util/listQueryBuilder.js
woyuen: This `getMinRangeValue` function was recently introduced to restrict returned results depending on their creation date, accordingly to compression/retention log policy for time-series tables.
It is called when the filter option `minValue` is provided.

https://github.com/stelace/stelace/blob/c2609113b4bf4280041c2c5bb62b9ff799a49210/src/services/event.js#L253-L259

Current list endpoint won't use it, except for the event list endpoint.

The purpose of this function is only to get the min value of the filter value or range. The throw is performed outside this function if the value doesn't match this condition:
https://github.com/stelace/stelace/blob/c2609113b4bf4280041c2c5bb62b9ff799a49210/src/util/listQueryBuilder.js#L113

Concerning the merging, it's to handle the special case where `gt` and `gte` is provided at the same time.

Example:
`/events?createdDate[gt]=2020-01-01T00:00:00.000Z&createdDate[gte]=2020-03-05T00:00:00.000Z`
which will be parsed as
```js
createdDate = {
  gt: '2020-01-01T00:00:00.000Z',
  gte: '2020-03-05T00:00:00.000Z'
}

getMinRangeValue(createdDate) === '2020-01-01T00:00:00.000Z'
```

So if the retention limit date is '2020-02-13T00:00:00.000Z', then the request will throw an error because results can only be filtered from '2020-02-13T00:00:00.000Z', not '2020-01-01T00:00:00.000Z'.

This function needs a proper comment on the usage.

src/util/listQueryBuilder.js
woyuen: Right, I'll add the full query for each case.

src/util/listQueryBuilder.js
woyuen: A comment was needed to be clear.

When using the continuous aggregate, `count` is a column value which is pre-aggregated, so we can use it directly.
In the other case, `count(*)` (equivalent to `knex.count()`) must be performed on the fly.

I'll add a comment for that.

woyuen:
I don't know why I can't reply to this individual comment, so I do it here.

> The problem is you probably omitted to mention you were destroying knex to avoid memory leaks by use of createSchema and dropSchema without destroying knex when not using. Am I right?
> 
> In such case, maybe the fix was just to rename destroyKnex (that was true by default) into returnKnex (false by default).
> 
> Anyway, you can reuse this commit instead of creating a new one, just explain your plan here.

Yes that's right, destroying knex instance by default was to ensure there's no omission after function call (especially in the case knex is created within the function).

Ok for renaming `destroyKnex` into `returnKnex`.

src/util/listQueryBuilder.js
abruere: You’re just partly rephrasing what I said, but you’re not providing an answer.

I compared with all current list endpoints where you can pass `gt[e]` and `lt[e]` applying to `createdDate` too (with implicit AND merging between `gt` and `lt` for instance).

This is why I was talking about inconsistency. Please read my message above again.

**EDIT**: also discussed on slack lately. Here is a summary for reference:

When using both `gte` and `gt` operators (`createdDate[gte]=2020-01-01&createdDate[gt]=2020-01-31`), @woyuen  you said `getMinRangeValue` takes the lowest value (equivalent to `OR` between two date comparisons) to check it’s higher than `minValue`, but you take the highest of `gte` and `gt` (equivalent to `AND`) when writing the SQL query as in `WHERE "createdDate" >= '2020-01-01' AND "createdDate" > '2020-01-31'`.

We agreed to ~remove~ refactor inconsistent `getMinRangeValue` and throw an error when redundant filters are used with this new history endpoint, `gte` and `gt`, according to principle of least suprise. And favoring explicit over implicit.

For all list endpoints, we’ll have to create another PR with a breaking change to make this consistent as well (no redundant use of `gte+gt` or `lte+lt`).

test/integration/api/event.spec.js
```diff
@@ -118,6 +118,32 @@ test('get events history with date filter', async (t) => {
   })
 })
 
+test('fails to get events history with same date boundary type filters', async (t) => {
```
abruere: Maybe `redundant relational operators` is more explicit.

Same for your commit message `Same boundary in range filter…` -> `Redundant relational operators`

src/util/validation.js
```diff
@@ -98,13 +98,19 @@ const objectIdParamsSchema = customJoi.object().keys({
   id: customJoi.string().required()
 }).required()
 
-function getRangeFilter (joiType) {
+function getRangeFilter (joiType, forbidSameBoundaryType = false) {
```
abruere, line 101: I suppose you will remove this parameter and make `true` the default behavior when migrating current list endpoints.

test/integration/api/event.spec.js
```diff
@@ -118,6 +118,32 @@ test('get events history with date filter', async (t) => {
...
+    .expect(400)
+
+  t.true(error1.message.includes('optional exclusive peers'))
+
+  const { body: error2 } = await request(t.context.serverUrl)
+    .get(`/events/history?groupBy=day&createdDate[lt]=${date}&createdDate[lte]=${date}`)
+    .set(authorizationHeaders)
+    .expect(400)
+
+  t.true(error2.message.includes('optional exclusive peers'))
```
abruere, line 144: Can you share the full Joi message here as a Github comment for reference?

abruere (commented):
Thanks!

It seems tests reached timeout again despite several attempts of yours after timestamptz change.

src/util/validation.js
woyuen: Exactly

test/integration/api/event.spec.js
woyuen: Sure. Here's the error message:
`"createdDate" contains a conflict between optional exclusive peers [gt, gte]`

docs/migration-to-timescaledb.md
```diff
@@ -0,0 +1,58 @@
...
+### From scratch
+
+If you don't need to keep existing data (e.g. in development environment) or can afford to start from scratch, please execute the following commands after checking your dev branch is up-to-date:
+
+```sh
+yarn docker:db:reset
+yarn docker:db
+```
+
+Now, the TimescaleDB container replaced PostgreSQL container.
```
abruere: Suggestion: `You’ve just replaced PostgreSQL container with TimescaleDB container.`

docs/migration-to-timescaledb.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Existing data
+
+To migrate data from a PostgreSQL database to a TimescaleDB database, Stelace provides a script (gist URL) you can use.
+
+*Please note this script is suitable if your time-series tables (event, webhookLog, workflowLog) have around 1 million rows each. Otherwise, you may need to adapt the script to import/export those tables in CSV format as explained in the [official documentation](https://docs.timescale.com/latest/getting-started/migrating-data). It will be a lot faster than transforming PostgreSQL tables into TimescaleDB hypertables.*
```
abruere: `have around 1 million rows each. ` -> `don’t have more than about 1 million rows each`

docs/migration-to-timescaledb.md
```diff
@@ -0,0 +1,58 @@
...
+
+Please copy the file into the root path of this project. Add the following environment variables into the file `.env`:
+
+```sh
+# Fill the TimescaleDB credentials
+TIMESCALE_HOST=
+TIMESCALE_PORT=
+TIMESCALE_DB=
+TIMESCALE_USER=
+TIMESCALE_PASSWORD=
```
abruere, line 45: By the way don’t you think we could keep existing `POSTGRES_*` env variable names since TimescaleDB is a mere extension of PostgreSQL?

woyuen: Ok, it's not clear enough. That's totally true.

Those environment variables only serve for the migration script. I miss to write the final step to replace the database credentials. I'll update the documentation, thanks.

abruere (approved):
Thank you for this PR!

When squashing commits, please remove redundant/patch commit messages from squashed commit message, so that we only keep helpful and documenting commit messages on dev/master branches (and full history here anyway).

PR #365, opened 2020-07-28
refactor!: rework workflow types for clarity

woyuen:
Previously there was only three types of workflow log:
- notification
- action
- runError

Any information concerning skipping or stopping the workflow step
was embedded into workflow log step object.

New types 'skipped' and 'stopped' are introduced to avoid checking
into this nested object.

Moreover, multiple variables had to be checked to determine
the cause of a log error. Now we distinguish two types of errors:
- preRunError (e.g. incorrect expressions), renaming from old runError
- "new" runError, replacing action with step.error = true

If the property `notifyUrl` of a workflow object is specified,
the object sent to the URL will have some property changes:
- `status` removed
- `type` added

### DIFF
--- a/src/services/workflow.js
+++ b/src/services/workflow.js
@@ -94,11 +94,25 @@ function start ({ communication, serverPort }) {
     if (populateLogs) {
       const queryBuilder = WorkflowLog.query().where({ workflowId: workflow.id })
       const actionCountQueryBuilder = queryBuilder.clone()
-        .whereIn('type', ['action'])
-        .whereJsonNotSupersetOf('step', { stopped: true })
-        .whereJsonNotSupersetOf('step', { skipped: true })
-        .whereJsonNotSupersetOf('step', { error: true, handleErrors: false })
-      const startedACountQueryBuilder = queryBuilder.clone().whereIn('type', ['action'])
+        .where(qb => {
+          return qb
+            .where(qb2 => {
+              return qb2
+                .whereIn('type', ['action'])
+
+                // DEPRECATED: old step object properties
+                .whereJsonNotSupersetOf('step', { stopped: true })
+                .whereJsonNotSupersetOf('step', { skipped: true })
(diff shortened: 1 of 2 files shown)


src/services/workflow.js
```diff
@@ -94,11 +94,27 @@ function start ({ communication, serverPort }) {
...
+                // DEPRECATED:END
+            })
+            .orWhere(qb2 => {
+              return qb2
+                .where('type', 'runError')
+                .whereJsonSupersetOf('step', { handleErrors: true })
+            })
+        })
+      const startedACountQueryBuilder = queryBuilder.clone().whereIn('type',
+        ['action', 'skipped', 'stopped', 'runError']
```
abruere: Why not everything _but_ `notification` type to ensure we don’t forget anything?

src/services/workflow.js
```diff
@@ -957,9 +974,10 @@ async function notifyAfterCompleted ({
 }) {
   const payload = {
     event: exposedEvent,
-    // if last step was successful, stopped/filtered or just skipped, this means the workflow was successful.
-    // An error interrupts the workflow.
-    status: lastLog.step.error ? 'error' : 'success',
```
abruere, line 962: Thanks for removing this redundant property of `notification` steps after discussion on slack.

Can you make sure it’s mentioned in commit message as well?

abruere (approved):
Looks good to me, thanks!

Please prepare [breaking change log (empty for now)](https://docs.api.stelace.com/?version=latest#7f785b07-7a40-4019-b571-473239ab7b8b) for API reference before merging, and include changes in future version draft.

src/services/workflow.js
woyuen: Indeed there are many types. It is everything but `notification` and `preRunError`. I'll update with a `whereNotIn` with those two types, thanks.

src/services/workflow.js
woyuen: Yes, sure !

PR #371, opened 2020-08-03
perf!: paginated endpoints with cursor pagination

woyuen:
Offset pagination isn't scalable for two reasons:
- retrieving count (full scan without caching strategy)
- offsetting results (scan first pages to reach wanted page)

Cursor pagination solves that issue by using cursors to directly
point to the location of the previous results limits.

Internally, this is transformed into basic comparisons,
which is really fast with indexes.

The drawback is that count isn't available anymore
and there is no way to directly jump into a specific page.

BREAKING CHANGE:
For version 2019-05-20, the offset pagination is served.
The cursor pagination is served for higher versions.

New platforms will directly use the new pagination.
But existing platforms will preserve the old pagination.

A manual system config update on `stelace.stelaceVersion`
to set a higher version is needed to use the new pagination globally.

Otherwise you can provide the header 'x-stelace-version'
at each request to check if everything works fine before migrating
globally.

### DIFF
--- a/plugins/rating/test/rating.spec.js
+++ b/plugins/rating/test/rating.spec.js
@@ -9,7 +9,11 @@ const {
 
 const { before, beforeEach, after } = lifecycle
 const { getAccessTokenHeaders } = auth
-const { checkStatsObject, checkOffsetPaginationScenario } = util
+const {
+  checkOffsetPaginationScenario,
+  checkOffsetPaginatedListObject,
+  checkOffsetPaginatedStatsObject,
+} = util
 
 test.before(async t => {
   await before({ name: 'rating' })(t)
@@ -53,7 +57,7 @@ test.serial('gets simple rating stats', async (t) => {
     .set(authorizationHeaders)
     .expect(200)
 
-  checkStatsObject({
+  checkOffsetPaginatedStatsObject({
     t,
     obj,
     groupBy,
@@ -107,7 +111,7 @@ test.serial('gets aggregated rating stats with ranking', async (t) => {
 
   let ranking
 
-  checkStatsObject({
+  checkOffsetPaginatedStatsObject({
     t,
     obj,
     groupBy,
@@ -204,18 +208,12 @@ test.serial('lists ratings with pagination', async (t) => {
 test('lists ratings with id filter', async (t) => {
   const authorizationHeaders = await getAccessTokenHeaders({ t, permissions: ['rating:list:all'] })
 
-  const result = await request(t.context.serverUrl)
+  const { body: obj } = await request(t.context.serverUrl)
     .get('/ratings?id=rtg_2l7fQps1I3a1gJYz2I3a')
     .set(authorizationHeaders)
     .expect(200)
 
-  const obj = result.body
-
-  t.is(typeof obj, 'object')
-  t.is(obj.nbResults, 1)
-  t.is(obj.nbPages, 1)
-  t.is(obj.page, 1)
-  t.is(typeof obj.nbResultsPerPage, 'number')
+  checkOffsetPaginatedListObject(t, obj)
   t.is(obj.results.length, 1)
 })
 
@@ -275,45 +273,28 @@ test('lists ratings with advanced filter', async (t) => {
 test('lists ratings with label filter', async (t) => {
   const authorizationHeaders = await getAccessTokenHeaders({ t, permissions: ['rating:list:all'] })
 
-  const result = await request(t.context.serverUrl)
+  const { body: obj } = await request(t.context.serverUrl)
     .get('/ratings?label=main:friendliness,main:pricing')
     .set(authorizationHeaders)
     .expect(200)
 
-  const obj = result.body
-
-  t.true(typeof obj === 'object')
-  t.true(typeof obj.nbResults === 'number')
-  t.true(typeof obj.nbPages === 'number')
-  t.true(typeof obj.page === 'number')
-  t.true(typeof obj.nbResultsPerPage === 'number')
-  t.true(Array.isArray(obj.results))
-
-  obj.results.forEach(rating => {
+  const checkResultsFn = (t, rating) => {
     t.true(['main:friendliness', 'main:pricing'].includes(rating.label))
-  })
+  }
+
+  checkOffsetPaginatedListObject(t, obj, { checkResultsFn })
 })
 
 test('lists ratings with wildcard label filter', async (t) => {
   const authorizationHeaders = await getAccessTokenHeaders({ t, permissions: ['rating:list:all'] })
 
-  const result = await request(t.context.serverUrl)
+  const { body: obj } = await request(t.context.serverUrl)
     .get('/ratings?label=main:*')
     .set(authorizationHeaders)
     .expect(200)
 
-  const obj = result.body
-
-  t.true(typeof obj === 'object')
-  t.true(typeof obj.nbResults === 'number')
-  t.true(typeof obj.nbPages === 'number')
-  t.true(typeof obj.page === 'number')
-  t.true(typeof obj.nbResultsPerPage === 'number')
-  t.true(Array.isArray(obj.results))
-
-  obj.results.forEach(rating => {
-    t.true(rating.label.startsWith('main:'))
-  })
+  const checkResultsFn = (t, rating) => t.true(rating.label.startsWith('main:'))
+  checkOffsetPaginatedListObject(t, obj, { checkResultsFn })
 })
 
 test('finds a rating', async (t) => {
--- a/plugins/rating/test/rating.spec.js
+++ b/plugins/rating/test/rating.spec.js
@@ -9,7 +9,7 @@ const {
 
 const { before, beforeEach, after } = lifecycle
 const { getAccessTokenHeaders } = auth
-const { checkStatsObject } = util
+const { checkStatsObject, checkOffsetPaginationScenario } = util
 
 test.before(async t => {
   await before({ name: 'rating' })(t)
@@ -17,6 +17,18 @@ test.before(async t => {
 })
 // test.beforeEach(beforeEach()) // Concurrent tests are much faster
 test.after(after())
+
+// need serial to ensure there is no insertion/deletion during pagination scenario
+test.serial('gets simple rating stats with pagination', async (t) => {
+  const authorizationHeaders = await getAccessTokenHeaders({ t, permissions: ['rating:stats:all'] })
+
+  await checkOffsetPaginationScenario({
+    t,
+    endpointUrl: '/ratings/stats?groupBy=authorId',
+    authorizationHeaders,
+    orderBy: 'avg'
+  })
+})
 
 // run this test serially because there is no filter and some other tests create events
 // that can turn the check on `count` property incorrect
@@ -178,22 +190,15 @@ test('gets aggregated rating stats with multiple labels', async (t) => {
   checkStatObject(results[0]['main:pricing'])
 })
 
-test('lists ratings', async (t) => {
+// need serial to ensure there is no insertion/deletion during pagination scenario
+test.serial('lists ratings with pagination', async (t) => {
   const authorizationHeaders = await getAccessTokenHeaders({ t, permissions: ['rating:list:all'] })
 
-  const result = await request(t.context.serverUrl)
-    .get('/ratings')
-    .set(authorizationHeaders)
-    .expect(200)
-
-  const obj = result.body
-
-  t.true(typeof obj === 'object')
-  t.true(typeof obj.nbResults === 'number')
-  t.true(typeof obj.nbPages === 'number')
-  t.true(typeof obj.page === 'number')
-  t.true(typeof obj.nbResultsPerPage === 'number')
-  t.true(Array.isArray(obj.results))
+  await checkOffsetPaginationScenario({
+    t,
+    endpointUrl: '/ratings',
+    authorizationHeaders,
+  })
 })
 
 test('lists ratings with id filter', async (t) => {
--- a/plugins/rating/test/rating.spec.js
+++ b/plugins/rating/test/rating.spec.js
@@ -13,6 +13,10 @@ const {
   checkOffsetPaginationScenario,
   checkOffsetPaginatedListObject,
   checkOffsetPaginatedStatsObject,
+
+  checkCursorPaginationScenario,
+  checkCursorPaginatedListObject,
+  checkCursorPaginatedStatsObject,
 } = util
 
 test.before(async t => {
@@ -26,7 +30,7 @@ test.after(after())
 test.serial('gets simple rating stats with pagination', async (t) => {
   const authorizationHeaders = await getAccessTokenHeaders({ t, permissions: ['rating:stats:all'] })
 
-  await checkOffsetPaginationScenario({
+  await checkCursorPaginationScenario({
     t,
     endpointUrl: '/ratings/stats?groupBy=authorId',
     authorizationHeaders,
@@ -57,7 +61,7 @@ test.serial('gets simple rating stats', async (t) => {
     .set(authorizationHeaders)
     .expect(200)
 
-  checkOffsetPaginatedStatsObject({
+  checkCursorPaginatedStatsObject({
     t,
     obj,
     groupBy,
@@ -111,7 +115,7 @@ test.serial('gets aggregated rating stats with ranking', async (t) => {
 
   let ranking
 
-  checkOffsetPaginatedStatsObject({
+  checkCursorPaginatedStatsObject({
     t,
     obj,
     groupBy,
@@ -125,7 +129,7 @@ test.serial('gets aggregated rating stats with ranking', async (t) => {
       t.is(typeof result.ranking, 'number')
       t.is(typeof result.lowestRanking, 'number')
 
-      t.is(result.lowestRanking, obj.nbResults) // is true because there is no filter
+      t.is(result.lowestRanking, obj.results.length) // is true because there is no filter
 
       // check ranking order
       if (typeof ranking === 'undefined') {
@@ -198,7 +202,7 @@ test('gets aggregated rating stats with multiple labels', async (t) => {
 test.serial('lists ratings with pagination', async (t) => {
   const authorizationHeaders = await getAccessTokenHeaders({ t, permissions: ['rating:list:all'] })
 
-  await checkOffsetPaginationScenario({
+  await checkCursorPaginationScenario({
     t,
     endpointUrl: '/ratings',
     authorizationHeaders,
@@ -213,7 +217,7 @@ test('lists ratings with id filter', async (t) => {
     .set(authorizationHeaders)
     .expect(200)
 
-  checkOffsetPaginatedListObject(t, obj)
+  checkCursorPaginatedListObject(t, obj)
   t.is(obj.results.length, 1)
 })
 
@@ -227,7 +231,6 @@ test('lists ratings with advanced filter', async (t) => {
 
   const obj1 = result1.body
 
-  t.is(obj1.results.length, obj1.nbResults)
   obj1.results.forEach(rating => {
     t.true(['usr_WHlfQps1I3a1gJYz2I3a', 'user-external-id'].includes(rating.authorId))
   })
@@ -239,7 +242,6 @@ test('lists ratings with advanced filter', async (t) => {
 
   const obj2 = result2.body
 
-  t.is(obj2.results.length, obj2.nbResults)
   obj2.results.forEach(rating => {
     t.true(['usr_WHlfQps1I3a1gJYz2I3a', 'user-external-id'].includes(rating.authorId))
   })
@@ -251,7 +253,6 @@ test('lists ratings with advanced filter', async (t) => {
 
   const obj3 = result3.body
 
-  t.is(obj3.results.length, obj3.nbResults)
   obj3.results.forEach(rating => {
     t.true(['usr_T2VfQps1I3a1gJYz2I3a'].includes(rating.targetId))
   })
@@ -263,7 +264,6 @@ test('lists ratings with advanced filter', async (t) => {
 
   const obj4 = result4.body
 
-  t.is(obj4.results.length, obj4.nbResults)
   obj4.results.forEach(rating => {
     t.true(['ast_0TYM7rs1OwP1gQRuCOwP'].includes(rating.assetId))
     t.true(['trn_UG1fQps1I3a1gJYz2I3a'].includes(rating.transactionId))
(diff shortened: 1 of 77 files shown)


src/util/pagination.js
```diff
@@ -14,8 +14,46 @@ function getPaginationMeta ({ nbResults, page, nbResultsPerPage }) {
...
+  order,
+  nbResultsPerPage,
+  page,
+  applyOrder = true,
+}) {
+  // Clone the query builder to have the count for all matched results before pagination filtering
+  const countQueryBuilder = queryBuilder.clone()
+
+  if (applyOrder) {
+    queryBuilder.orderBy(orderBy, order)
```
abruere, line 58: Shouldn’t we call `orderBy` if `orderBy` and `order` are provided rather than using an additional parameter?

I don’t see a place where you compute this `applyOrder` boolean parameter dynamically which would justify its presence.

test/util.js
```diff
@@ -109,6 +114,110 @@ async function testEventMetadata ({
...
+  }
+
+  const orders = ['desc', 'asc']
+
+  for (const order of orders) {
+    let orderValue
+    const checkOrder = (results) => {
+      results.forEach(r => {
+        if (!_.isUndefined(orderValue)) {
+          if (order === 'desc') t.true(r[orderBy] <= orderValue)
```
abruere: I’d rename `orderValue` to `previousValue` (this is much like a *reduce*).

test/util.js
```diff
@@ -109,6 +114,110 @@ async function testEventMetadata ({
...
+  t.true(_.isNumber(obj.nbPages))
+  t.true(_.isNumber(obj.page))
+  t.true(_.isNumber(obj.nbResultsPerPage))
+  t.true(Array.isArray(obj.results))
+
+  if (obj.nbPages === 1) t.is(obj.results.length, obj.nbResults)
+  if (_.isFunction(checkResultsFn)) obj.results.forEach(r => checkResultsFn(t, r))
+}
+
+async function checkOffsetPaginationScenario ({
```
abruere, line 248: Thanks for this util function.

Don’t you think you should check result pages are not overlapping? (some of the last results of previous page repeated in current page).

plugins/rating/test/rating.spec.js
```diff
@@ -53,7 +57,7 @@ test.serial('gets simple rating stats', async (t) => {
     .set(authorizationHeaders)
     .expect(200)
 
-  checkStatsObject({
+  checkOffsetPaginatedStatsObject({
```
abruere: I don’t think you need to rename these calls. Why not checking the format of pagination in current `checkStatsObject` rather than having dedicated function? (offset pagination in this commit, cursor based pagination in next commits).

src/util/cursor.js
```diff
@@ -0,0 +1,93 @@
+const _ = require('lodash')
+const { encodeBase64, decodeBase64 } = require('./encoding')
+
+const allowedTypes = [
+  'number',
+  'boolean',
+  'string',
+
+  // Date instances are returned from database for timestamptz type
```
abruere: Do you mean this `date` type refers to timestamptz?

To be more explicit:

> This refers to timestamptz dates returned from PostgreSQL
> which is different from date stored in string type

src/util/cursor.js
```diff
@@ -0,0 +1,93 @@
...
+  // which is different from date stored in string type
+  'date',
+]
+
+// must be uncommon chain of characters to distinguish
+// from real string values
+const separator = '~|~'
```
abruere, line 16: To avoid this, did you think about (base64-)encoding a stringified JSON object?

Then you wouldn’t need a separator (and could have more than 2 values for unicity but probably YAGNI)

Just an idea, please don’t rush into this.

src/util/cursor.js
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * A cursor encodes one or two values.
+ * If unicity isn't guaranteed by one value, provide two values.
+ * Example: (createdDate) isn't enough because two rows can have the same value
+ * then (createdDate, id) is a good candidate for cursor
+ * @param {Object}   obj - cursor will be created based on this object
+ * @param {Object[]} config
+ * @param {Object}   config[i]
+ * @param {String}   config[i].prop - object property to encode
```
abruere: Maybe you should rename config to `props` and `prop` property of each array item to `name`.

`props` parameter name probably more explicit than `config`. Same for `areValidCursorProps` instead of `isValidCursorConfig`.

What do you think?

src/util/pagination.js
```diff
@@ -1,3 +1,33 @@
+const _ = require('lodash')
+const createError = require('http-errors')
+const { createCursor, parseCursor } = require('./cursor')
+
+const reverseOrder = (order) => order === 'asc' ? 'desc' : 'asc'
+const reverseOperator = (operator) => {
+  // reverse the order, not negate the operator
+  // negation of '>' is '<='
```
abruere: You could add parentheses to this comment to make it clear it’s an irrelevant example.

src/util/pagination.js
```diff
@@ -51,9 +81,177 @@ async function offsetPaginate ({
...
+/**
+ * @param {Object}   queryBuilder - Knex.js query builder
+ *
+ * `startingAfter` and `endingBefore` are mutually exclusive
+ * @param {String}   [startingAfter] - if specified, fetch results after this cursor
+ * @param {String}   [endingBefore] - if specified, fetch results before this cursor
+ *
+ * @param {Number}   nbResultsPerPage
+ *
+ * @param {Object[]} cursorConfig - will be used to create/parse cursor
```
abruere: Same: maybe `cursorProps`.

src/util/pagination.js
```diff
@@ -51,9 +81,177 @@ async function offsetPaginate ({
...
+// ///////////////// //
+// CURSOR PAGINATION //
+// ///////////////// //
+
+/**
+ * @param {Object}   queryBuilder - Knex.js query builder
+ *
+ * `startingAfter` and `endingBefore` are mutually exclusive
+ * @param {String}   [startingAfter] - if specified, fetch results after this cursor
+ * @param {String}   [endingBefore] - if specified, fetch results before this cursor
```
abruere: > fetch **last** results

src/util/pagination.js
```diff
@@ -51,9 +81,177 @@ async function offsetPaginate ({
...
 
+// ///////////////// //
+// CURSOR PAGINATION //
+// ///////////////// //
+
+/**
+ * @param {Object}   queryBuilder - Knex.js query builder
+ *
+ * `startingAfter` and `endingBefore` are mutually exclusive
+ * @param {String}   [startingAfter] - if specified, fetch results after this cursor
```
abruere: > fetch **first** results

src/util/pagination.js
```diff
@@ -51,9 +81,177 @@ async function offsetPaginate ({
...
+  order,
+}) {
+  // if `endingBefore` is specified, this is equivalent to retrieving
+  // the `nbResultsPerPage` last results.
+  // To achieve it, this is done with 2 steps:
+  // 1. reverse the `order` direction for the SQL query with limit to `nbResultsPerPage`
+  // 2. reverse the obtained results in Javascript
+  const shouldReverseOrder = !!endingBefore
+
+  const cursor = startingAfter || endingBefore
```
abruere, line 120: As in previous PRs:

Don’t you think we should only allow one of these? For least surprise, rather than silently picking one.

By the way, in a subsequent commit you make it exclusive https://github.com/stelace/stelace/pull/371/commits/044fabbbdefd1995afdc9c657d67d47de8cbc7a3#diff-285f19a9ab4a0e60e77a40695003d5ebR50

src/util/pagination.js
```diff
@@ -51,9 +81,177 @@ async function offsetPaginate ({
...
+  const paginationMeta = getCursorPaginationMeta({
+    decodedCursor,
+    results,
+    startingAfter,
+    endingBefore,
+    nbResultsPerPage,
+  })
+
+  // remove cursor result
+  if (decodedCursor) results = results.filter(_.negate(matchCursorResult(decodedCursor)))
```
abruere, line 158: Nice functional style.

src/util/pagination.js
```diff
@@ -1,3 +1,33 @@
...
+    '>': '<',
+    '>=': '<=',
+    '<': '>',
+    '<=': '>=',
+  }
+
+  return map[operator]
+}
+
+const matchCursorResult = _.curry((decodedCursor, result) => {
```
abruere, line 19: Good idea to use curry for functional use. :+1:

src/util/pagination.js
```diff
@@ -51,9 +81,177 @@ async function offsetPaginate ({
...
+        qb.where(firstProp, operatorWithEqual, decodedCursor[firstProp])
+      } else {
+        qb
+          .where(firstProp, operator, decodedCursor[firstProp])
+          .orWhere(qb2 => {
+            // if two-columns cursor
+            // use operator with equal on the secondary column to include the cursor result
+            // to determine if there is a previous page
+            return qb2
+              .where(firstProp, decodedCursor[firstProp])
```
abruere, line 243: It’s not clear in `createCursor` function above that the strict `order` only applies to second property when two of them are provided, as a **tie breaker**.

This would explain why you only use `operatorWithEqual` with second property (and simple equality operator with first prop).

src/util/math.js
```diff
@@ -1,3 +1,21 @@
+const Big = require('big.js')
+
+// Classic native JS sum has floating precision issues:
+// Basic sum: 8.1 + 8.2 = 16.299999999999997
+
+// Transitive issue:
+// 8.3 + 8.7 + 8 + 8.1 = 33.1
+// 8 + 8.7 + 8.1 + 8.3 = 33.099999999999994
```
abruere, line 8: :rofl: 

Thanks for taking care of this. Which tests made you notice?

src/versions/validation/index.js
```diff
@@ -112,7 +112,11 @@ const middleware = () => {
       return setImmediate(next)
     }
 
-    const { schema } = validationDefinition
+    const { schema: schemaObjOrFunction } = validationDefinition
```
abruere, line 115: In your commit message, you say:

> This solves definition objects order in validation files.

Could you elaborate with an example?

test/auth.js
```diff
@@ -19,12 +20,18 @@ async function getAccessTokenHeaders ({
...
-  return {
+  const headers = {
     authorization: `Bearer ${accessToken}`,
     // TODO: use an API key instead of these special headers
     'x-platform-id': t.context.platformId,
     'x-stelace-env': t.context.env
   }
+
+  if (apiVersion) {
+    headers['x-stelace-version'] = apiVersion
```
abruere, line 32: For reference, can you explain why you need this?
It seems we can already ask for a specific pagination on a per-request basis.

Why do you need this to get access tokens?
Probably for SDK and pagination handling, but context would be useful to future reader anyway.

plugins/rating/test/rating.spec.js
```diff
@@ -125,7 +129,7 @@ test.serial('gets aggregated rating stats with ranking', async (t) => {
       t.is(typeof result.ranking, 'number')
       t.is(typeof result.lowestRanking, 'number')
 
-      t.is(result.lowestRanking, obj.nbResults) // is true because there is no filter
+      t.is(result.lowestRanking, obj.results.length) // is true because there is no filter
```
abruere, line 132: Don’t forget than removing `nbResults` is a also one of the breaking changes.

Did we make this choice to avoid redundancy and even confusion with total number of results?

plugins/rating/versions/validation/rating.js
```diff
@@ -22,6 +22,31 @@ module.exports = function createValidation (deps) {
...
+  // ////////// //
+  schemas['2020-08-10'] = {}
+  schemas['2020-08-10'].getStats = () => ({
+    query: schemas['2019-05-20'].getStats.query
+      .fork('page', schema => schema.forbidden())
+      .keys({
+        // cursor pagination
+        startingAfter: Joi.string(),
+        endingBefore: Joi.string(),
+      })
```
abruere: We should probably refactor this in a shared object (including `oxor` below), except if it’s too tricky with our use of Joi.

src/versions/request/beforeAll.js
```diff
@@ -0,0 +1,17 @@
+const requestChanges = {
+  '2019-05-20': [
+    {
+      target: 'beforeAll', // all requests with a version <= '2019-05-20'
+      description: 'Use offset pagination for version below version 2019-05-20 included',
+      up: async (params) => {
+        const { req } = params
+
+        req._useOffsetPagination = true
```
abruere, line 9: Nice to have version transformers :smile:

src/communication.js
```diff
@@ -141,7 +141,10 @@ async function stelaceApiRequest (endpointUri, {
   const defaultHeaders = {
     'x-platform-id': platformId,
     'x-stelace-env': env,
-    'x-stelace-system-key': process.env.SYSTEM_KEY
+    'x-stelace-system-key': process.env.SYSTEM_KEY,
+
+    // force the new version to have cursor pagination
+    'x-stelace-version': '2020-08-10'
```
abruere, line 147: This may answer my previous question about access tokens.

abruere (commented):
Thanks!
First round.

src/util/pagination.js
woyuen: There is only one case where the order params must not be applied:
https://github.com/stelace/stelace/blob/0807210e6b8e4138c376586cdd667617d8ae4815/src/services/document.js#L330

This is because the document stats aggregation already performs it with the WITH CLAUSE queries.
I should have added a comment mentioning the use at that exact location.

test/util.js
woyuen: Exact! I will rename it.

test/util.js
woyuen: Good suggestion! I will implement that.

plugins/rating/test/rating.spec.js
woyuen: That's a good question.
I preferred separate the naming to force testing a specific type of pagination (offset for old version and cursor for the new one).
A `paginationType` parameter defaulting to 'cursor' may do the job. What do you think?

src/util/cursor.js
woyuen: Exactly. Thanks for the modified comment. I'll update it.

src/util/cursor.js
woyuen: Yes indeed. Your suggestion is a good one.
This separator might break if this exact value is used by users (even if it's unlikely).
The only drawback will be the cursor length but it may be worth.

src/util/cursor.js
woyuen: Good idea!

src/util/pagination.js
woyuen: I didn't make it exclusive in this util function because I relied on the validation step.
I'll add the exclusive parameter check here too.

src/util/math.js
woyuen: With pagination scenario testing, the cursor pagination rearranged the order differently due to 2-columns order.
And that average check was broken:
https://github.com/stelace/stelace/blob/6b5a6ef5619a71b4f7ca175c46c4883c83708067/test/util.js#L395

I suspected TimescaleDB number precision. It was Javascript in reality...

src/versions/validation/index.js
woyuen: Yes that wasn't clear enough.

I wanted to define validation schemas difference in anti-chronological order.
Without this modification, it wasn't possible to reference an object that is define further in the file.

Example: `schemas['2020-08-10'].list` references `schemas['2019-05-20'].list`
https://github.com/stelace/stelace/blob/6b5a6ef5619a71b4f7ca175c46c4883c83708067/src/versions/validation/event.js#L68-L78
https://github.com/stelace/stelace/blob/6b5a6ef5619a71b4f7ca175c46c4883c83708067/src/versions/validation/event.js#L113-L134

Now validation schema can be a function that returns the validation schema only when needed (at the endpoint hit).

test/auth.js
woyuen: Indeed, I'll add a comment for that.

The request header 'x-stelace-version' force the use of a specific API version even if the platform is in a different API version. Putting this parameter on `getAccessTokenHeaders` function is practical because it is used by almost all tests when setting API permissions.

By providing the parameter `apiVersion` in this function, tests can check that old behaviour (from older versions platforms) is still working. So the breaking change is correctly versioned.
In this case, I wanted to test that old platforms will work with offset pagination.

plugins/rating/test/rating.spec.js
woyuen: This `nbResults` property from pagination meta represents the total number of results (which is different from `nbResultsPerPage`) that won't be available with cursor pagination.

plugins/rating/versions/validation/rating.js
woyuen: Indeed, I'll make this change, thanks.

src/util/pagination.js
woyuen: Indeed, I'll improve that comment, thanks!

src/util/pagination.js
abruere: This is not what I meant: why add the `applyOrder` parameter when you can just see if `orderBy` and `order` are provided?

A parameter would only make sense if you wanted to compute it dynamically (when you don’t know in advance you have to provide `orderBy` and `order` just in case).

plugins/rating/test/rating.spec.js
abruere: If you want to keep explicit mention of pagination type, parameter has no real added value compared with different function names. You can keep current implementation.

test/auth.js
abruere: Then your comment should just stay this allows to skip `x-stelace-version` header in individual requests, right?

plugins/rating/test/rating.spec.js
abruere: Yes, it was confusing as it could also be mistakenly seen as the number of results of the current page (can be less than `nbResultsPerPage` in last page).

src/util/cursor.js
abruere: It’s an implementation detail hidden from interface, right (even when not using our JS SDK)? Which would mean no breaking change in the future (provided we detect and handle both formats).

In this case maybe we can keep this as it is for now and refactor if needed in the future.

src/util/math.js
abruere: Thanks, what the second/ordering column was?

src/versions/validation/index.js
abruere: Nice, this may be a performance hit though, unless you use singleton constants.

src/util/pagination.js
woyuen: Ok I misunderstood. No need to introduce this `applyOrder` then. Thanks!

src/util/cursor.js
woyuen: Yes indeed. We can change the cursor format in the future with backward compatibility.

src/util/math.js
woyuen: Oh wait, I think I made a mistake. It was the stats check function so there is no second column.
If I remembered correctly, it was the rating test (`avg desc`):
https://github.com/stelace/stelace/blob/6b5a6ef5619a71b4f7ca175c46c4883c83708067/plugins/rating/test/rating.spec.js#L642-L652

src/versions/validation/index.js
woyuen: Indeed, I'll update the code to avoid this performance issue.

test/auth.js
woyuen: Yes indeed.

plugins/rating/test/rating.spec.js
woyuen: Yes it is. This test rely on the relative few number of aggregated results (less than 20 results).
I'll add a comment for it.

abruere: No comment needed.

The confusing part I’m referring to is the deprecated `nbResults`.

src/util/cursor.js
```diff
@@ -30,43 +30,43 @@ function isValidCursorConfig (config) {
...
+ * @param {Object[]} props
+ * @param {Object}   props[i]
+ * @param {String}   props[i].name - prop name to encode
+ * @param {String}   props[i].type - allowed value: 'number', 'boolean', 'date', 'string'
  */
-function createCursor (obj, config) {
-  if (!config) throw new Error('Missing cursor config')
-  if (!isValidCursorConfig(config)) throw new Error('Invalid cursor config')
+function createCursor (obj, props) {
+  if (!props) throw new Error('Missing cursor props')
```
abruere, line 39: Maybe full word `properties` is better in error messages intended for users.

src/util/cursor.js
```diff
@@ -30,43 +30,43 @@ function isValidCursorConfig (config) {
...
+ * @param {Object}   props[i]
+ * @param {String}   props[i].name - object property
+ * @param {String}   props[i].type - allowed value: 'number', 'boolean', 'date', 'string'
  */
-function parseCursor (cursor, config) {
-  if (!config) throw new Error('Missing cursor config')
-  if (!isValidCursorConfig(config)) throw new Error('Invalid cursor config')
+function parseCursor (cursor, props) {
+  if (!props) throw new Error('Missing cursor props')
+  if (!isValidCursorProps(props)) throw new Error('Invalid cursor props')
```
abruere, line 59: Same.

test/auth.js
abruere: Your added comment is confusing.

I wrote

> allows to skip `x-stelace-version` header in individual requests

because we already had this option before. What you want to do when including it in access token is to skip this header in each and every individual request using this token, right?
In this case this is not what your comment says.

src/util/cursor.js
woyuen: Actually this error is intended to be displayed to us as API designers, because it warns us that the `cursorProps` passed by [`performListQuery`](https://github.com/stelace/stelace/blob/3880051b4d30aea2bcbf9146c61636544428e20f/src/util/listQueryBuilder.js#L295-L298) or [`performHistoryQuery`](https://github.com/stelace/stelace/blob/3880051b4d30aea2bcbf9146c61636544428e20f/src/util/listQueryBuilder.js#L645-L647) has an incorrect format.

This error will never be displayed to API consumers. It will be revealed thanks to failing cursor pagination test.

woyuen:
> Your added comment is confusing.

Indeed, I should have added your comment. I'll update it.

src/util/cursor.js
abruere: Does this also applies to other occurences?

abruere (approved):

src/util/cursor.js
woyuen: Yes it does.

PR #379, opened 2020-08-14
feat: add SSL options to PostgreSQL connection

woyuen:
Some database providers only accepts requests with SSL enabled
(e.g. Timescale Cloud).

We support the following SSL options: ssl, cert, key, ca
Those options will map from Knex.js to the library
https://github.com/iceddev/pg-connection-string
thanks to the Knex.js PR
https://github.com/knex/knex/pull/3410

Please be careful when configuring SSL connection.
Passing key value into cert or inversely can trigger
an uncaught exception and even `process.on('uncaughtException', fn)`
cannot catch it.

Add multi-domains certificate to PostgreSQL Docker image to check
that SSL connections is correctly handled in tests.


docker-compose.override.yml
```diff
@@ -35,6 +35,22 @@ services:
     volumes:
       - timescaledb-v1.7.1-pg12-data:/var/lib/postgresql/data/pgdata
 
+  # Service used only test environment to check if SSL database connections work
```
abruere: > ~**in** test environment~

EDIT: removed after refactoring.

abruere (approved):
Thank you @woyuen.

Please make sure appropriate context of your individual commit messages and of this PR are included in squashed commit message before merging.

PR #382, opened 2020-08-15
feat: add history endpoint for workflow logs

woyuen:
Only system can request this endpoint for now


abruere (approved):
Thank you, looks good to me.

PR #389, opened 2020-08-20
fix!: some issues on list filters

woyuen:
A test util is implemented to automatically check all filters
specified in a config, with values fetched from
fixtures or manually provided.

Three types of filters are tested:
- exact value filter (`prop=value`)
- array filter (`prop[]=value1&prop[]=value2` or `prop=value1,value2`)
- range filter (`prop[gte]=value1&prop[lte]=value2`)

Some issues detected:
- array filter doesn't work with some endpoints (format: `prop=value1,value2`)
- range filter isn't applied when using single value
- unused filter

BREAKING CHANGE:
Filters `assetId` and `transactionId` from rating stats and list endpoints
don't accept array values anymore.
This is because document `data` filter used internally
to filter those properties doesn't
filter a single value property with an array on its
JSONB column.
Moreover, even if we adapt the `data` filter to accept array values
on some defined properties, it will transform the query into
`data->'assetId' <@ [id1,id2]::jsonb` which isn't indexed by
current GIN index.

### DIFF
--- a/plugins/rating/test/rating.spec.js
+++ b/plugins/rating/test/rating.spec.js
@@ -2,6 +2,7 @@ require('dotenv').config()
 
 const test = require('ava')
 const request = require('supertest')
+const _ = require('lodash')
 
 const {
   testTools: { lifecycle, auth, util }
@@ -17,6 +18,8 @@ const {
   checkCursorPaginationScenario,
   checkCursorPaginatedListObject,
   checkCursorPaginatedStatsObject,
+
+  checkFilters,
 } = util
 
 test.before(async t => {
@@ -198,6 +201,79 @@ test('gets aggregated rating stats with multiple labels', async (t) => {
   checkStatObject(results[0]['main:pricing'])
 })
 
+// use serial because no changes must be made during the check
+test.serial('check history filters', async (t) => {
+  const authorizationHeaders = await getAccessTokenHeaders({
+    t,
+    permissions: [
+      'rating:stats:all',
+      'rating:list:all',
+    ]
+  })
+
+  const groupBys = [
+    'authorId', // document first-level property
+    'assetId', // document data sub-property
+  ]
+
+  for (const groupBy of groupBys) {
+    const { body: { results } } = await request(t.context.serverUrl)
+      .get('/ratings?nbResultsPerPage=100')
+      .set(authorizationHeaders)
+      .expect(200)
+
+    const resultsByGroupBy = _.groupBy(results, groupBy)
+
+    const customExactValueCheck = _.curry((prop, obj, value) => {
+      let filteredResults = resultsByGroupBy[obj[groupBy]] || []
+      filteredResults = filteredResults.filter(r => r[prop] === value)
+      return filteredResults.length === obj.count
+    })
+
+    const customArrayValuesCheck = _.curry((prop, obj, values) => {
+      let filteredResults = resultsByGroupBy[obj[groupBy]] || []
+      filteredResults = filteredResults.filter(r => values.includes(r[prop]))
+      return filteredResults.length === obj.count
+    })
+
+    await checkFilters({
+      t,
+      endpointUrl: `/ratings/stats?groupBy=${groupBy}`,
+      fetchEndpointUrl: '/ratings',
+      authorizationHeaders,
+      checkPaginationObject: checkCursorPaginatedListObject,
+
+      filters: [
+        {
+          prop: 'authorId',
+          isArrayFilter: true,
+          customCheck: customArrayValuesCheck('authorId'),
+        },
+        {
+          prop: 'targetId',
+          isArrayFilter: true,
+          customCheck: customArrayValuesCheck('targetId'),
+        },
+        {
+          prop: 'assetId',
+          // isArrayFilter: true,
+          customCheck: customExactValueCheck('assetId'),
+        },
+        {
+          prop: 'transactionId',
+          // isArrayFilter: true,
+          customCheck: customExactValueCheck('transactionId'),
+        },
+        {
+          prop: 'label',
+          // multiple labels filter on stats are tested on other tests
+          customCheck: customExactValueCheck('label'),
+        },
+      ],
+    })
+  }
+})
+
 // need serial to ensure there is no insertion/deletion during pagination scenario
 test.serial('lists ratings with pagination', async (t) => {
   const authorizationHeaders = await getAccessTokenHeaders({ t, permissions: ['rating:list:all'] })
@@ -219,6 +295,43 @@ test('lists ratings with id filter', async (t) => {
 
   checkCursorPaginatedListObject(t, obj)
   t.is(obj.results.length, 1)
+})
+
+// use serial because no changes must be made during the check
+test.serial('check list filters', async (t) => {
+  const authorizationHeaders = await getAccessTokenHeaders({ t, permissions: ['rating:list:all'] })
+
+  await checkFilters({
+    t,
+    endpointUrl: '/ratings',
+    authorizationHeaders,
+    checkPaginationObject: checkCursorPaginatedListObject,
+
+    filters: [
+      {
+        prop: 'id',
+        isArrayFilter: true,
+      },
+      {
+        prop: 'authorId',
+        isArrayFilter: true,
+      },
+      {
+        prop: 'targetId',
+        isArrayFilter: true,
+      },
+      {
+        prop: 'assetId',
+      },
+      {
+        prop: 'transactionId',
+      },
+      {
+        prop: 'label',
+        isArrayFilter: true,
+      },
+    ],
+  })
 })
 
 test('lists ratings with advanced filter', async (t) => {
--- a/plugins/rating/test/rating.spec.js
+++ b/plugins/rating/test/rating.spec.js
@@ -246,13 +246,11 @@ test.serial('check history filters', async (t) => {
       filters: [
         {
           prop: 'authorId',
-          isArrayFilter: true,
           customExactValueFilterCheck: customExactValueCheck('authorId'),
           customArrayFilterCheck: customArrayValuesCheck('authorId'),
         },
         {
           prop: 'targetId',
-          isArrayFilter: true,
           customExactValueFilterCheck: customExactValueCheck('targetId'),
           customArrayFilterCheck: customArrayValuesCheck('targetId'),
         },
--- a/plugins/rating/test/rating.spec.js
+++ b/plugins/rating/test/rating.spec.js
@@ -247,27 +247,27 @@ test.serial('check history filters', async (t) => {
         {
           prop: 'authorId',
           isArrayFilter: true,
-          customCheck: customArrayValuesCheck('authorId'),
+          customExactValueFilterCheck: customExactValueCheck('authorId'),
+          customArrayFilterCheck: customArrayValuesCheck('authorId'),
         },
         {
           prop: 'targetId',
           isArrayFilter: true,
-          customCheck: customArrayValuesCheck('targetId'),
+          customExactValueFilterCheck: customExactValueCheck('targetId'),
+          customArrayFilterCheck: customArrayValuesCheck('targetId'),
         },
         {
           prop: 'assetId',
-          // isArrayFilter: true,
-          customCheck: customExactValueCheck('assetId'),
+          customExactValueFilterCheck: customExactValueCheck('assetId'),
         },
         {
           prop: 'transactionId',
-          // isArrayFilter: true,
-          customCheck: customExactValueCheck('transactionId'),
+          customExactValueFilterCheck: customExactValueCheck('transactionId'),
         },
         {
           prop: 'label',
           // multiple labels filter on stats are tested on other tests
-          customCheck: customExactValueCheck('label'),
+          customExactValueFilterCheck: customExactValueCheck('label'),
         },
       ],
     })
--- a/test/integration/api/event.spec.js
+++ b/test/integration/api/event.spec.js
@@ -2,6 +2,7 @@ require('dotenv').config()
 
 const test = require('ava')
 const request = require('supertest')
+const _ = require('lodash')
 
 const { before, beforeEach, after } = require('../../lifecycle')
 const { getAccessTokenHeaders } = require('../../auth')
@@ -14,7 +15,11 @@ const {
   checkCursorPaginatedHistoryObject,
   checkCursorPaginationScenario,
   checkCursorPaginatedListObject,
+
+  checkFilters,
 } = require('../../util')
+
+const { truncateDate } = require('../../../src/util/time')
 
 test.before(async t => {
   await before({ name: 'event' })(t)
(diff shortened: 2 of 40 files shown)


test/util.js
```diff
@@ -513,3 +515,261 @@ function checkCursorPaginatedHistoryObject ({
...
+    addCheckObjFunctions(filter)
+
+    const getValue = filter.getValue
+
+    const useCustomValues = customTestValues && customTestValues.length
+
+    const realValues = _.compact(_.uniq(results.map(getValue)))
+    let testValues = useCustomValues
+      ? customTestValues
+      : _.shuffle(realValues).slice(0, 2) // only take few values
```
abruere: Why only 2? Takes so much time?

test/util.js
```diff
@@ -513,3 +515,261 @@ function checkCursorPaginatedHistoryObject ({
...
+      customTestValues,
+    } = filter
+
+    addCheckObjFunctions(filter)
+
+    const getValue = filter.getValue
+
+    const useCustomValues = customTestValues && customTestValues.length
+
+    const realValues = _.compact(_.uniq(results.map(getValue)))
```
abruere: The scope for `results` seems to be implicit (strong code smell).

This whole `checkFilters` function is quite contrieved and the vague JS doc comment doesn’t help.

test/util.js
```diff
@@ -513,3 +515,261 @@ function checkCursorPaginatedHistoryObject ({
...
+      if (testValues.length === 1) {
+        const randomValue = getRandomValue(sortedValues)
+
+        filter.testRangeValues = {
+          gte: randomValue,
+          lte: randomValue,
+        }
+      } else {
+        filter.testRangeValues = {
+          gte: _.first(sortedValues),
```
abruere: Are you sorting just to use `_.first`?

Sorting add complexity and doesn’t really help with performance compared to a plain `every` check.

test/util.js
```diff
@@ -513,3 +515,261 @@ function checkCursorPaginatedHistoryObject ({
...
+
+          // filter only on the first half set of results to check if the range filter is effective
```
abruere: Why?

test/util.js
```diff
@@ -513,3 +515,261 @@ function checkCursorPaginatedHistoryObject ({
...
+      customGetValue,
+      isArrayFilter,
+      isRangeFilter,
+      customCheck
+    } = filter
+
+    const getValue = (obj) => _.isFunction(customGetValue) ? customGetValue(obj) : obj[prop]
+    filter.getValue = getValue
+
+    filter.getExactValueCheckObj = _.curry((exactValue, obj) => {
```
abruere: It seems to be `get…CheckFn` rather than `get…CheckObj`.

Same for the other functions below.

test/util.js
```diff
@@ -513,3 +515,261 @@ function checkCursorPaginatedHistoryObject ({
...
+ * @param {String}   [params.fetchEndpointUrl] - if not specified, fallback to endpointUrl
+ *   is the endpoint to hit to auto-generate test values
+ *   (useful to specify it for stats and history endpoints)
+ * @param {Function} [params.checkPaginationObject(t, obj)]
+ * @param {Object[]} params.filters
+ * @param {String}   params.filters[i].prop
+ * @param {Function} [params.filters[i].customGetValue(obj)]
+ *   custom getter to value with prop (by default obj[prop])
+ * @param {Function} [params.filters[i].customCheck(obj, value)]
+ *   if specified, will perform a custom check on that property
```
abruere: It would be helpful to state which checks are performed by default.

test/util.js
```diff
@@ -513,3 +515,261 @@ function checkCursorPaginatedHistoryObject ({
...
+    const getValue = filter.getValue
+
+    const useCustomValues = customTestValues && customTestValues.length
+
+    const realValues = _.compact(_.uniq(results.map(getValue)))
+    let testValues = useCustomValues
+      ? customTestValues
+      : _.shuffle(realValues).slice(0, 2) // only take few values
+
+    if (!testValues.length) testValues = ['value1', 'value2']
```
abruere: Why is this needed?

It’s confusing and error-prone since it can make the test useless.

plugins/rating/test/rating.spec.js
```diff
@@ -198,6 +201,79 @@ test('gets aggregated rating stats with multiple labels', async (t) => {
...
+          customCheck: customArrayValuesCheck('authorId'),
+        },
+        {
+          prop: 'targetId',
+          isArrayFilter: true,
+          customCheck: customArrayValuesCheck('targetId'),
+        },
+        {
+          prop: 'assetId',
+          // isArrayFilter: true,
```
abruere: Why do you keep these comments with `isArrayFilter` set to true (it’s false by default).

abruere (commented):
Thanks.

It seems your new `checkFilter` may need some refactoring or more proper use:

When you use `isArrayFilter: true` and don’t provide any value to test (relying on some default placeholders), it doesn’t really test anything except that the endpoint exists and accept array arguments, not that they’re working properly.

test/util.js
woyuen: I restrict to only two different values because some fixtures don't have more than 2 different values for the specified filter.
Without restricting the values enough, it's like requesting all values with the array filter and we cannot be sure if the filter works or not.

`values.includes(getValue(obj))` would be similar to `allValues.includes(getValue(obj))` and the condition will always be true.
https://github.com/stelace/stelace/blob/47a66145b591f0c5e35d0dd3ac5c5be1107d1ba4/test/util.js#L602-L606

test/util.js
woyuen: Thanks for the catch. `results` must be passed as parameter to the function `addTestValues`. This change was needed after some refactor when implementing the util function.

I'll add more details on how to use this util and how it should be have.

test/util.js
woyuen: Yes the sorting is to prepare the values for the range filter. That's to ensure we pass coherent values for `gte` and `lte` (no `lte` value inferior than `gte` value, which will leads to empty results).
You're right. I'll remove the sorting and just get random values and pick min/max values for respective bounds.

test/util.js
woyuen: This is for the same reason as array filter. We filter only a subset of available values (half of the interval is arbitrary) to ensure the range filter works.

The condition `rangeValues.gte <= getValue(obj) && getValue(obj) <= rangeValues.lte` will check that the returned results have the expected values.
https://github.com/stelace/stelace/blob/47a66145b591f0c5e35d0dd3ac5c5be1107d1ba4/test/util.js#L608-L612

test/util.js
woyuen: Yes thanks!

test/util.js
woyuen: Indeed, I'll add explanations on that.

plugins/rating/test/rating.spec.js
woyuen: Thanks for the catch. The comments has to be removed after removing rating array filter.

test/util.js
woyuen: Those fake values was to ensure the array filter format is allowed, in the case there aren't any values retrieved from fixtures.
This is indeed confusing, I'll remove it.

test/util.js
```diff
@@ -517,6 +517,19 @@ function checkCursorPaginatedHistoryObject ({
 }
 
 /**
+ * Util to check list/stats endpoint filters works as expected
+ * 3 types of filters is supported:
```
abruere: *are* supported

test/util.js
```diff
@@ -517,6 +517,19 @@ function checkCursorPaginatedHistoryObject ({
...
+ * - exact value filter
+ * - array filter
+ * - range filter
+ *
+ * If test values aren't provided, values will be automatically fetched from fixtures.
+ *
+ * Each filter will be checked that the validation format is correct
```
abruere: Typos/grammar

test/util.js
```diff
@@ -529,8 +542,19 @@ function checkCursorPaginatedHistoryObject ({
  * @param {String}   params.filters[i].prop
  * @param {Function} [params.filters[i].customGetValue(obj)]
  *   custom getter to value with prop (by default obj[prop])
- * @param {Function} [params.filters[i].customCheck(obj, value)]
- *   if specified, will perform a custom check on that property
+ *
+ * Custom check filter functions, one by filter type
+ * `getValue` is the getter function (obj[prop] or customGetValue(obj))
```
abruere: **the result of** the getter function?

test/util.js
```diff
@@ -529,8 +542,19 @@ function checkCursorPaginatedHistoryObject ({
...
+ * @param {Function} [params.filters[i].customExactValueFilterCheck(obj, value)]
+ *  where value is a primitive type
+ *  default: (obj, value) => getValue(obj) === exactValue
+ * @param {Function} [params.filters[i].customArrayFilterCheck(obj, values)]
+ *  where values is an array
+ *  default: (obj, value) => values.includes(getValue(obj))
+ * @param {Function} [params.filters[i].customRangeFilterCheck(obj, rangeValues)]
+ *  where rangeValues is an object with the following properties: `gte` and `lte`
+ *  default: rangeValues.gte <= getValue(obj) && getValue(obj) <= rangeValues.lte
```
abruere: `gt` and `lte` are supported in some endpoints but not here?

test/util.js
```diff
@@ -529,8 +542,19 @@ function checkCursorPaginatedHistoryObject ({
...
+ *
  * @param {Boolean}  [params.filters[i].isArrayFilter = false]
```
abruere: Do you still need this now that you have three types of custom check functions, including a specific ones for array and range?

woyuen: Yes I still need `isArrayFilter` and `isRangeFilter` because some endpoints don't need custom check functions, so we need to differentiate them.

test/util.js
woyuen: That's right. I tested only `gte` and `lte` operators but not all combinations.
I'll add them too to cover all cases, thanks.

test/util.js
abruere: Should not they be optional when providing corresponding custom check functions?

woyuen: Yes indeed, I make them optional if custom check functions are provided.

test/integration/api/event.spec.js
```diff
@@ -184,9 +184,12 @@ test.serial('check history filters', async (t) => {
     return filteredResults.length === obj.count
   })
 
-  const customRangeValuesCheck = _.curry((prop, obj, rangeValues) => {
+  const customRangeValuesCheck = _.curry((prop, obj, rangeValues, isWithinRange) => {
     let filteredResults = results.filter(r => truncateDate(r.createdDate) === obj.day)
-    filteredResults = filteredResults.filter(r => rangeValues.gte <= r[prop] && r[prop] <= rangeValues.lte)
+    filteredResults = filteredResults.filter(r => isWithinRange(
+      prop === 'createdDate' ? obj.day : r[prop],
```
abruere: This `day` property` applies to history endpoint, right?

It seems that rounding convention is not specified in API reference draft.

test/integration/api/role.spec.js
```diff
@@ -31,7 +31,7 @@ test.serial('lists roles', async (t) => {
 })
 
 // use serial because no changes must be made during the check
-test.serial('check list filters', async (t) => {
+test.serial.only('check list filters', async (t) => {
```
abruere: This prevents other tests from being run.

test/integration/api/event.spec.js
woyuen: Yes indeed. The property `day` is the value of the `groupBy`. I should have added a `groupBy` variable so it would be explicit.

I'll explain the rounding convention in the API reference draft.

abruere (approved):

test/util.js
abruere: > We filter only a subset of available values (half of the interval is arbitrary) to ensure the range filter works.

This still does not explain why. You’re merely rephrasing your current comment I was asking about.

~And `rangeValues.gte <= getValue(obj) && getValue(obj) <= rangeValues.lte` looks like a typo to me (do you really want `<= lte`)?~

test/util.js
abruere: > I restrict to only two different values because some fixtures don't have more than 2 different values for the specified filter.

Real examples would be more helpful.

test/util.js
woyuen: I'll add an example for clarity.
Let's suppose we have a number filter on resources whose values spreads from 1 to 10.
If we didn't pick a subset of the interval, the filter will apply the range `[1, 10]`, which is equivalent to the absence of the filter in this case because we filter all results. By filtering the range `[1, 5]` and checking that results match the filter, we ensure it works.
In the last version of the code, I picked random values (if there are more than one value) so it isn't half of the interval.
https://github.com/stelace/stelace/blob/3941883f5e2bf591034f4e4ac97693269ff44526/test/util.js#L679-L698

Concerning the condition `rangeValues.gte <= getValue(obj) && getValue(obj) <= rangeValues.lte`, it seems to me there isn't a typo. This function was refactored in a subsequent commit to allow all range combinations. Maybe it's clearer this way.
https://github.com/stelace/stelace/blob/3941883f5e2bf591034f4e4ac97693269ff44526/test/util.js#L600-L609

test/util.js
woyuen: Yes indeed. For instance, webhook logs statuses can only have two values `success` or `error`.
https://github.com/stelace/stelace/blob/dcea4e38bed6462058fa11e10f6df92ddd1d9463/src/services/webhook.js#L437-L443

Maybe I can add a parameter to `checkFilters` to specify the number of values to use when retrieving automatically from fixtures to handle that special case and that won't affect all array filters, so we can increase the default value from 2 to 5.

abruere: It’s still not clear why the fixtures should not only include filter values handled by the API endpoint, such as `success` and `error` for webhook log `status`?

test/util.js
abruere: > If we didn't pick a subset of the interval, the filter will apply the range [1, 10], which is equivalent to the absence of the filter in this case because we filter all results. By filtering the range [1, 5] and checking that results match the filter, we ensure it works.

This is the piece of simple explanation I was looking for, thanks.

> Concerning the condition `rangeValues.gte <= getValue(obj) && getValue(obj) <= rangeValues.lte`, it seems to me there isn't a typo.

My bad, I’ve read too fast and I was confused with the inversion of operands.

test/util.js
woyuen: I meant the webhook log fixtures will only include `success` and `error` as `status` values, those are handled by the API. However in this special case where there are very few different allowed values, checking with a high number of filter values will be equivalent of the absence of the filter in the test. That's why I only took 2 values by default.

But a better solution might to apply specify `customTestValues: ['error', 'success']` for this case so the `checkFilter` function will individually check those values one by one.

If we increase the number of filter values from 2 to 5, we must use `customTestValues` for each filter whose the number of different values allowed by API is inferior to 5.

abruere: Thanks for the explanation. The problem with your random approach is that tests may or may not fail depending on tested filter values, when they are expected to fail due to some regression.

Explicit is better than implicit. But you may have good reasons to only pick some filter values, if there are hundreds of them and you don’t want to test them all because there is no reason to have a regression for a particular value.

This is clearly not the case for discrete values with specific behaviors to test such as webhook log `status`.

abruere (approved):
Looks ready for merge after ensuring that:

- API reference updates are ready
- Front-end demos are updated accordingly if necessary (offending filters refactored, or if impossible, removed before next deployment)
- your PR description is more precise for future reference. Maybe related to `data` filter using `jsonSupersetOf` ? https://github.com/stelace/stelace/blob/60bfbf07122a11c79e93249d351a62101cda25ef/src/services/document.js#L432-L435

> This is because document data filter used internally to filter those properties doesn't filter a single value property with an array on its SONB column.

with -> with**in**?

PR #399, opened 2020-09-01
refactor: improve test robustness

woyuen:
- remove order from condition to prevent failing tests due to parallel execution in workflow test suite
- add logs filter to better isolate from logs generated by other tests
- prevent database connection terminated error by creating a new test platform

### DIFF
--- a/test/lifecycle.js
+++ b/test/lifecycle.js
@@ -155,38 +155,47 @@ function before ({ name, platformId, env, enableSignal = true, useFreePort } = {
 
 function beforeEach ({ minimumFixtures = false } = {}) {
   const fn = async (t) => {
-    const systemKey = getSystemKey()
-
     const platformId = t.context.platformId
 
     for (const env of testingEnvs) {
-      await request
-        .post(`${t.context.serverUrl}/store/platforms/${platformId}/database/drop`)
-        .set(getAuthorizationHeaders({ systemKey, env }))
-        .catch(handleDropDatabaseError)
-
-      await request
-        .post(`${t.context.serverUrl}/store/platforms/${platformId}/database/migrate`)
-        .set(getAuthorizationHeaders({ systemKey, env }))
-
-      let fixtures = getDataFixtures(env)
-
-      if (minimumFixtures) {
-        fixtures = _.pick(fixtures, [
-          'config',
-          'roles',
-          'user'
-        ])
-      }
-
-      const connection = getPostgresqlConnection({ platformId, env })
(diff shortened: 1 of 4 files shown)


test/lifecycle.js
```diff
@@ -197,6 +206,38 @@ function after () {
   return fn
 }
 
+async function createPlatform ({ t, minimumFixtures = false }) {
```
abruere, line 209: Before refactoring, how were new platforms created in tests (just to make sure you can’t use existing logic).

By the way, did you make a typo in your commit message?

> tests are **not** total**ly** isolated

abruere (approved):

test/lifecycle.js
woyuen: Before refactoring, there was only the test suite `store.spec.js` that created new platforms in tests.
https://github.com/stelace/stelace/blob/22b27b56bd68e337043e9d5cfbba25097e2966b5/test/integration/api/store.spec.js#L252-L255
https://github.com/stelace/stelace/blob/22b27b56bd68e337043e9d5cfbba25097e2966b5/test/integration/api/store.spec.js#L363-L372

However the platform was only created with those requests. The configuration part was missing (databases credentials, databases start, seeding). That wasn't a problem in `store.spec.js` file because the test was to check the platform was correctly initialized.

The objective of the function `createPlatform` I introduced in `lifecycle.js` is to return a fully functional platform.
This logic was present but was split between `before()` and `beforeEach()` lifecycle functions, so I refactored it into `createPlatform` function.

> By the way, did you make a typo in your commit message?
> tests are not totally isolated

Yes thanks, I made a typo.

[EDIT]: I realize the commit message isn't clear enough as it uses the word `re-create` in `those files needed to re-create the test platform to have fresh fixtures`. That can be understood as we create a new platform from scratch while we reuse existing platform by dropping data and initializing again in reality. I propose the following message:

```
This error can happen in `asset.spec.js` and `assetType.spec.js` files
because those files needed to reused the test platform created
during lifecycle function `.before()` by dropping data and initializing again
to have fresh fixtures.
However even using AVA `.serial()`, tests aren't totally isolated
and some database connections are dropped while still processing.
```

What do you think?

abruere: > However the platform was only created with those requests.

I was refering to lifecycle. You obviously need to create new platforms (with some config) to run tests. This was the shareable logic I’m talking about.

> What do you think?

OK. It seems there is a typo: `needed to reused`

woyuen: Thanks. I'll merge the PR when the test passes (commit message updated).

PR #512, opened 2020-12-08
Minor bugfixes for docker build process

lynsei:
- The Ava dependency was not there, causing an error when testing.  I think this might have been a global installed CLI that somebody forgot to put into the package.json since it wasn't visible.
- Also the dotenv module was throwing an error I installed it as a dev dependency and it fixed itself, but I could not reproduce it after.  I'm pretty sure it was not the NPM dependency, but the configuration of the docker test commands, which I have now changed a bit.
-  I'm not using your Circle CI, so I won't know if this works in your environment.  But I have used Circle CI before, and I think it should be okay ;)
- I added some documentation about the node version.  If it were me, I'd remove the local builds entirely because docker builds are way more stable.  I'd also at a Makefile or build.sh file to handle the environment and docker configurations necessary.  Ideally you want your docker builds to be deterministic, so using the "build" directive is not recommended.  Instead I'd say pre-create all the database images in a single Dockerfile using a multi-stage build (multiple FROM statements).  ->  Push them to Dockerhub.  That way you greatly reduce the docker-compose time, and things are semantically versioned.  
- You can also load your env files into docker like this: `docker-compose --env-file ./config/.env.dev up` 
- 12 Factor apps say to not have different sets of env files or (docker-compose) files for different environments. I'd say it would be better to use a Makefile and scale up that way.

Just my 2 cents 👍


package.json
```diff
@@ -32,14 +32,14 @@
     "plugins": "npm run plugins:install && npm run plugins:prepare",
     "plugins:install": "node scripts/installExternalPlugins.js",
     "plugins:prepare": "node scripts/prepareInstalledPlugins.js",
-    "test": "NODE_ENV=test ava --c $(node -p 'Math.max(os.cpus().length - 2, 1)')",
-    "test:server": "cp -n .env.example .env && yarn && NODE_ENV=test ava --c $(node -p 'Math.max(os.cpus().length - 2, 1)')",
```
abruere, line 36: `cp -n .env.example .env` is not only there for users who may have forgotten to create their `.env` file, but as a default env for plugins too when running server tests after self-loading.

abruere (commented):
Thank you for all your suggestions !
The project build process needs some refactoring indeed, and would benefit from using `FROM` statements and a Makefile.

I agree it’s confusing to have Docker deployment not being default (forcing users to make some changes in the env file as stated in the README), as it is usually not advised to use database containers instead of proper databases in production. But as you said production deployment configuration should belong to its own files. We’ll probably improve this in the near future.

**As for ava**:
Can you share the sequence of commands that failed before your PR and your OS, so we can reproduce ?
I don’t have ava installed globally and I can run ava-related commands without issues on Linux. `ava-cli` is a 5-year old fix that should not be necessary anymore with current version of ava.

PR #540, opened 2021-01-13
Update README: plugins before seeding & use master

justinschuldt:
Three things here:
- I found that running `yarn seed` before running `yarn plugins` does not work, so I updated the order of the commands in the README.
- Fixed the link to the yarn installation docs, it was pointing to a `404`. I guess [they don't have complete translations](https://github.com/yarnpkg/website/issues/993) anymore after switching the way they do i18n.
- updated the `git clone ...` command to checkout `master` after cloning, so that people following the guide get the stable branch by default.

p.s. nice codebase and documentation!

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,12 +108,12 @@ Choose the right installation guide for your system.
 
 ### Install yarn
 
-Please refer to [official instructions](https://yarnpkg.com/fr/docs/install).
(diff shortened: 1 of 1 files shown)


abruere:
Thank you so much !

(By the way we shouldn’t have used the French URL for yarn docs anyway.)

